PR #2, opened 2018-11-14
updated repo with new cookiecutter

huard:
The test suite does not work.


cehbrecht:
... the indices processes are not activated.

huard:
Correct. Coming up later ; )

PR #4, opened 2018-11-14
Xclim test

huard:
## Overview

There is now a basic test for the Xclim based indicator. 

One issue is that pytest leaves a symbolic link to a file created during testing...


cehbrecht:
I have fixed the dependencies (add xclim from git and removed eggshell). Made a local test with `finch.tx_max` and it worked :) 

I haven't seen the left-over link by pytest .... Please run `make clean install` to get a clean conda env.

cehbrecht:
@huard Python 2.7 tests are failing due a unicode error:
`UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 38: ordinal not in range(128)`

I suppose the string comes from the indices description ... I haven't checked.

huard:
The UnicodeError is due to docstrings in xclim missing the "u". 
See https://github.com/Ouranosinc/xclim/pull/108
That should fix the Py2 problem.

PR #8, opened 2018-12-11
Docker fixes

davidcaron:
## Overview

Changes:

- Don't copy the code before installing conda dependencies. This way the dependencies are cached by docker and subsequent builds are faster.
- Install gunicorn and psycopg2 to use in a production environment
- The default command starts gunicorn, but the development server can still be started by using: `docker run -p 5000:5000 finch finch [options] start`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,19 @@
 .git
 docs/
 tests/
+src/
+finch.egg-info/
+
+# Any pywps generated log files
+**/*.log
+
+# Unit test / Coverage reports
+.cache
+.pytest_cache
+
+# PyCharm
+*.idea
+
+# Byte-compiled / optimized / DLL files
+**/__pycache__
+**/*.py[cod]
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,27 +5,32 @@ LABEL Description="Finch WPS" Vendor="Birdhouse" Version="0.1.0"
 
 # Update Debian system
 RUN apt-get update && apt-get install -y \
- build-essential \
-&& rm -rf /var/lib/apt/lists/*
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
 
 # Update conda
 RUN conda update -n base conda
 
-# Copy WPS project
-COPY . /opt/wps
-
 WORKDIR /opt/wps
 
 # Create conda environment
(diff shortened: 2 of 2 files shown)


davidcaron:
@cehbrecht, I thought about adding `conf.d` to the Dockerfile. This way we could configure the container with environment variables only. 

This blog post is a good explanation: http://www.mricho.com/confd-and-docker-separating-config-and-code-for-containers/
And this is the github repo: https://github.com/kelseyhightower/confd

Right now, when I start a container I mount a config file to replace `finch/default.cfg`. 
If we use `conf.d`, we install it along with a template file in a file in `finch/templates` in the Dockerfile. This template file contains `{{ getenv "EXAMPLE" }}` directives that are replaced when the container is started.

I wanted to know what you thought about this.

davidcaron:
Nevermind, I think I found a simpler solution. When I run the container, I use the PYWPS_CFG environment variable. So my `docker-compose.yml` looks like:
``` yaml
version: '3'
services:
  finch:
    image: finch
    command: ["gunicorn", "--config", "/code/gunicorn.py", "finch.wsgi:application"]
    environment:
      - PYWPS_CFG=/wps.cfg
    volumes:
      - ./config/finch/gunicorn.py:/code/gunicorn.py
      - ./config/finch/wps.cfg:/wps.cfg
      - wpsoutputs:/wpsoutputs
```

Also, we don't need to create a folder for wps outputs in the Dockerfile at build time, as it's created by the bind mount when the container starts.

I think this makes the Dockerfile much simpler.

davidcaron:
> Shouldn't the docker image have a `pywps` config that works for the default settings for this container? To provide your own `wps.cfg` via environment variable would be optional then.

Yes, the `default.cfg` file is also loaded. So these are the defaults, and anything in `PYWPS_CFG` should override these defaults.

Regarding nginx, I tried it and it works well.

I think it would make sense to have another repo similar to https://github.com/bird-house/ansible-wps-playbook but for a docker deployment. I'm not yet at a stage where I can recommend my solution as a 'best practice'.

huard:
@davidcaron Is it ready to merge ?

davidcaron:
It's deployed and working at finch.crim.ca so yes, you can merge this. I"ll add another PR if necessary.

PR #10, opened 2019-03-05
Development

davidcaron:
## Overview

Multiple fixes and improvements, I tried to split the commits for each. The commit messages should explain the changes.

Also, xarray dropped support for python 2.7, so we don't have a choice but to drop it too.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,6 @@
+.dockerignore
+Dockerfile
+
 .git
 docs/
 tests/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.7"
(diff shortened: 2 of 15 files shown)


huard:
@davidcaron `sentry_sdk` is listed in requirements_prod, but not in `environment.yml`. I've added it directly in master. 

Also, I see that you needed to clean the `_data` attribute otherwise it ends up in the database. I guess this is something we should fix in PyWPS, no ? Was that part of the PR you made ?

davidcaron:
@huard, I had to check to make sure, but yes this is fixed by my recent PR. The problem was that the _data property was written to the database when serializing a UrlHandler type, and now it's not. WRiting to the database was less a problem than returning the whole data as part of the status document.

Issue #36, opened 2019-05-21
Performance for subsetting a gridpoint is very slow


davidcaron:
## Description

Subsetting the BCCAQv2 datasets alog the time dimension takes 5-10 minutes for a single file. There are 270 files, so the processing is needlessy long for this simple operation.

While modifying the original data is not desired, here are the proposed solutions:

 - Don't do anything, keep using the original data
 - re-chunk the data
 - re-align the data so that the time dimension is last

Related to https://github.com/bird-house/finch/pull/33

tlogan2000:
I am creating a few re-chunked files for us to test - I will keep original packing but writing chunked to disk for both in space and time 
First validation will be to ensure new data is identical to original
- [ ] Chunked data is identical to original

```
# rechunk test
path_to_bccaqv2 = '/home/logan/boreas/pcic/BCCAQv2/'
outpath = '/home/logan/boreas/testdata/Chunking_Netcdf/bccaqv2_daily'
datasets = sorted(glob.glob(glob.os.path.join(path_to_bccaqv2,'*tasmin*.nc')))
print("starting")
ChunkSizes = [365, 50, 56]
comp = dict(chunksizes=ChunkSizes)
for dataset in datasets[0:2]: 
    t = time.perf_counter()
    # to keep original 'packing' use mask_and_scale=False 
    ds = xr.open_dataset(dataset,mask_and_scale=False,decode_times=False, chunks=dict(time=365))
    encoding = {var: comp for var in ds.data_vars} 
    encoding['time'] = dict(dtype='single')
    out_fn = glob.os.path.join(outpath , "ReChunk_Test_packed_{n}.nc".format(n=glob.os.path.basename(dataset)))
    ds.to_netcdf(out_fn,encoding=encoding)
    print(f"{time.perf_counter() - t:.2f} seconds")

tomLandry:
Ok, so I fill in with more info on our daily planning for this. The idea is to get a sample script today (seems you're working on this, much appreciated) so we can run it on one of our machines. That yields the baseline to compute throughput. We then scale the number of machines. Idea is to hit "play" on the "distributed" script on Friday, and come back on Monday with a shiny, new, speady dataset ready to use at CRIM.

tlogan2000:
On my end I am not sure exactly how to rewrite a transposed dataset (realigned option).  
@huard Do you have a feel as to what exactly the difference would be between this and re-chunking as time:-1, lon: 1, lat:1?

[...]

PR #37, opened 2019-05-21
[WIP] Output results as a csv instead of a netcdf

davidcaron:
## Overview

For subsetting and index processing, merge the datasets in a csv when requested.

Different datetime types will have to be tested.


davidcaron:
@tlogan2000, how should we merge the data in the different models and rcps?

If I do `xr.open_mfdataset("tasmin*.nc").to_dataframe().to_csv("subset_result/out.csv")` with multiple models, I get a csv that has one observation for each day, for each model, and for each rcp. so each day has 86 observations.

I see the following options:
 - A single csv file with the following columns: `time,lon,lat,tasmin,model,rcp`
 - one csv file per model and rcp with the columns: `time,lon,lat,tasmin`

tlogan2000:
I think this is where the different calendars are causing you problems.. I will get back on this as I got distracted with the performance issues

davidcaron:
When I merge using only `.to_dataframe().to_csv()` on each variable, it seems to work... 

I tested this with this dataset. This is a set of netcdf files for a single grid point, for every model and rcp, for the year 2016. [BCCAQv2_subset_45.385644_-73.691963.zip](https://github.com/bird-house/finch/files/3204143/BCCAQv2_subset_45.385644_-73.691963.zip)

tlogan2000:
This is more the type of output I would think end-users might expect.  Each columns is the data for an individual variable/model/rcp. Column names are changed to reflect each combination
For now am outputing 1 file per calendar type ...

```python
import xarray as xr
import glob
import os.path as path
import pandas as pd
import numpy as np
datafolder = '{path_to_data}/BCCAQv2_subset_45.385644_-73.691963'
outfolder = '{path_to_data}/BCCAQv2_subset_45.385644_-73.691963/output'
ncfiles = sorted(glob.glob(path.join(datafolder,'*tasmin*.nc')))
# get unique calendar types
calendars =[]
for n in ncfiles:
    ds = xr.open_dataset(n,decode_times=False)
    calendars.append(ds.time.calendar)

# intialize dict
csv_type = {}
for i in list(set(calendars)):
    csv_type[i] = []

for n, ncfile in enumerate(ncfiles):
    ds = xr.open_dataset(ncfile, decode_times=False)
    cal = ds.time.calendar
    ds['time'] = xr.decode_cf(ds).time
    for v in ds.data_vars:
        outvar = '{v}_{m}_{r}'.format(v=v,m=ds.attrs['driving_model_id'],r=ds.attrs['driving_experiment_id'].replace(',','_'))
        ds = ds.rename({v:outvar})
    df = ds.to_dataframe()
    df = df[['lat','lon',outvar]]
    
    # most runs have timestamp with hour == 12 a few hour == 0 .. make uniform 
    df.index = df.index.map(lambda x: x.replace(hour=12))

    if not isinstance(csv_type[cal],pd.DataFrame):
        csv_type[cal] = df.copy()

    else:
        csv_type[cal] = pd.concat([csv_type[cal], df[outvar]], axis=1)
# export to csv
for t in csv_type:
    outcsv = path.join(outfolder,'csv_out{t}.csv'.format(t=t))
    csv_type[t].to_csv(outcsv)

**Comment by huard** [comment]
@davidcaron See utils.py, I've added a few functions to print a header with global attributes and variable attributes. It's not connected yet to any process. 

@tlogan2000  If we mix models and RCPs in the same csv output, it's going to be complicated to include meta data. 

I agree there is value in making the dates and time uniform across simulations, though. Maybe we should store the metadata in independent files to avoid cluttering the main csv table.

**Comment by tomLandry** [comment]
If we have to split by any of these dimensions, I would split by RCP. It's the easiest to figure out for a user, on the UI part. It's a nice even split, selfcontained, no information is missing. The download API already covers it.

**Comment by tlogan2000** [comment]
If creating a separate metadata file is an option this would be a pretty simple solution adding a few lines of code in the for loop

```
for n, ncfile in enumerate(ncfiles):
    ## previous code here...

    ## metadata additional code
    metadata = pd.DataFrame.from_dict([ds.attrs]).transpose()
    metadata.columns = ['{name}:netcdf_metadata'.format(name=outvar)]
    if not isinstance(meta_all,pd.DataFrame):
        meta_all = metadata.copy()
    else:
        meta_all = pd.concat([meta_all, metadata], axis=1,sort=True)

# export metadata
meta_all.to_csv(path.join(outfolder,'bccaqv2_metadata.csv'))

huard:
You probably want variable metadata as well, if only for the units.

davidcaron:
Ok, so... from what I understand, here is what seems to stand out:

 - Having uniform calendar types is a good idea according to Travis and David H. So grouping by calendar type is pretty much inevitable.
 - We want header and variable metadata. This information take around 70 lines for each file.
 - If the files are grouped, it's hard to put the metadata in the same file. Well... not hard, but awkward.

So the metadata should be in a separate file. Whether it should be a single metadata file or multiple, I'm not sure, but it's not hard to change.

I'll go ahead and implement this, using David H. and Travis's code. Let me know If you have other thoughts on this.

tlogan2000:
code example with variable attributes added to the output metadata:

```
meta_all = []
for n, ncfile in enumerate(ncfiles):
    ds = xr.open_dataset(ncfile, decode_times=False)
    cal = ds.time.calendar
    ds['time'] = xr.decode_cf(ds).time
    for v in ds.data_vars:
        outvar = '{v}_{m}_{r}'.format(v=v,m=ds.attrs['driving_model_id'],r=ds.attrs['driving_experiment_id'].replace(',','_'))
        ds = ds.rename({v:outvar})
    df = ds.to_dataframe()
    df = df[['lat','lon',outvar]]



    # most runs have timestamp with hour == 12 a few hour == 0 .. make uniform
    df.index = df.index.map(lambda x: x.replace(hour=12))

    if not isinstance(csv_type[cal],pd.DataFrame):
        csv_type[cal] = df.copy()


    else:
        csv_type[cal] = pd.concat([csv_type[cal], df[outvar]], axis=1)

    metadata = pd.DataFrame.from_dict([ds.attrs]).transpose()
    metadata.columns = ['{name}:netcdf_metadata'.format(name=outvar)]
    var_meta = pd.DataFrame.from_dict([ds[outvar].attrs]).transpose()
    var_meta.columns = ['{name}:netcdf_metadata'.format(name=outvar)]
    metadata = pd.concat([var_meta, metadata], axis=0)
    if not isinstance(meta_all,pd.DataFrame):
        meta_all = metadata.copy()
    else:
        meta_all = pd.concat([meta_all, metadata], axis=1,sort=False)

meta_all.to_csv(path.join(outfolder,'model_metadata.csv'))

**Comment by davidcaron** [comment]
@tlogan2000 @huard Here is an example output if you want to take a look: 
[output.zip](https://github.com/bird-house/finch/files/3209317/output.zip)

The metadata is in its own folder, and there are 4 csv files, one for each calendar type.

**Comment by davidcaron** [comment]
One more thing: is it possible that processing indices doesn't keep some metadata? I think I can't find the `driving_model_id` in the output file after calculating indices.

**Comment by tlogan2000** [comment]
I think the indices operating only on the dataarray (and not the dataset) would end up losing the global attributes of the original bccaqv2 file

You could do something like the following to create an output dataset with original attrs
```
import xclim.atmos as atmos
ds1 = xr.open_dataset(bccaqv2_file)
# calc xclim index
out = atmos.tx_mean(ds1.tasmax)
# Create empty output dataset with original dims and attrs
dsOut = xr.Dataset(data_vars=None, coords=out.coords, attrs=ds1.attrs)
# Add our climate index 'out' as a data variable to the dataset
dsOut[out.name] = out
print(dsOut)

tlogan2000:
I think the output looks good..  I am not sure about whether it is best to have metadata all together or in separate files but but as long as it is there I think we're fine

For the actual data csvs i would maybe only suggest that we add the units directly to the column names to make it explicitly clear what users are downloading 
something like this should do the trick:
 
```
df = ds.to_dataframe() # previous code
df = df[['lat','lon',outvar]] #previous code to order columns
# add units to variable columns name
df.columns = ['lat','lon','{v} ({u})'.format(v = outvar, u=ds[outvar].units)] # new line

tomLandry:
Thanks @tlogan2000. In terms of daily priorities, once we are confortable with the CSV export, we tackle reorganisation of the dataset for speedy access. So chunking or reordering, while preserving precision (or with scale/offset). We plan to compute everything starting tomorrow, so new dataset would be ready for Monday. I'm getting a bit off-topic, so we can open another channel for this.

davidcaron:
Subsetting performance can be discussed in this issue: https://github.com/bird-house/finch/issues/36

davidcaron:
I added the original global attributes of the input dataset and the units to the columns names. I think this would be done. I'll test it a bit more locally before deploying.

> I am not sure about whether it is best to have metadata all together or in separate files but but as long as it is there I think we're fine

I agree it makes a lot of files... but I tend to think that regrouping that inside a single file makes it hard to find the dataset you're looking for. It's only my opinion, if anyone wants to change it, it's fine with me.

PR #39, opened 2019-05-27
Subset using threads and configure bccaqv2 url

davidcaron:
Add possibility to subset a gridpoint using multiple threads and configure the bccaqv2 opendap url in the .cfg file


davidcaron:
@huard I will merge this, but your review is still welcome if you have time :-)

huard:
Nothing to report.

PR #40, opened 2019-05-29
Development

davidcaron:
## Overview

 - In one of our datasets, the key `driving_model_id` is named `GCM__model_id`. I will investigate more why this is the case, and remove the code if the dataset gets changed
 - Multiple threads were trying to write the status document at the same time. This would cause the monitoring to fail when getting the status of the task. I added a lock.
 - Some retrys for flaky thredds connections
 - possibilty to use local files instead of a remote thredds server
 - temporary fix for a part of https://github.com/Ouranosinc/xclim/issues/213 (broken dataset, not heat wave computation time)


huard:
@davidcaron Were you planning to merge these ? If there are upstream fix to xclim we need to make, please let us know.

davidcaron:
Yes, sorry, I need to clean this up... a lot of these are related only to BCCAQv2 processes, but I still need to check and merge these.

huard:
No problem. I'm giving a refresh to Finch, updating it with the latest xclim indicators, etc. I don't want to get too far ahead and create conflicts with this PR though.

PR #47, opened 2019-11-19
bugfix: convert Path to str in MetaFile assignment in subset

### LINKED ISSUE #46: Fix bug with file path
## Description

Subset processes fail because MetaFile is given a Path instead of a file string. 

This should also be supported by PyWPS eventually.

huard:
## Overview

This PR fixes #46


tlvu:
Would like to release a new Finch docker image, can this PR be merged?

PR #57, opened 2019-12-17
Identify opendap by header content description

### LINKED ISSUE #23: More reliable way to identify opendap url
## Description

Currently, when we want to know if a url is an Opendap url, we append `.dds` to it and check if we get a result.

This causes a problem when sending a link of the following format:

.../thredds/dodsC/birdhouse/nrcan/nrcan_canada_daily_v2/pr/nrcan_canada_daily_pr_2017.nc?pr[0:1:3000][0:1:2][0:1:2]

where appending .dds to the url returns a 404.

We need a more resilient way to find if a url id an opendap url or not.

davidcaron:
## Overview

This PR fixes #23 
From the docstring of `is_opendap_url`:

>    The DAP Standard specifies that a specific tag must be included in the
    Content-Description header of every request. This tag is one of:
        "dods-dds" | "dods-das" | "dods-data" | "dods-error"
    So we can check if the header starts with `dods`.
    Even then, some OpenDAP servers seem to not include the specified header...
    So we need to let the netCDF4 library actually open the file.

## Additional Information

section 7.1.1 at: https://www.opendap.org/pdf/ESE-RFC-004v1.2.pdf

Also from the netcdf-java implementation: 

> Theres an ambiguity as to whether http://server/something is an OPeNDAP or an HTTP remote file using range requests. There will be more ambiguities in the future, as other HTTP based protocols are added. Currently we do a HEAD request on http://server/something.dds, and if it succeeds, and returns a header  Content-Description="dods-dds" or "dods_dds", then we open as OPeNDAP, and if it fails we try opening as an HTTP file. 

https://www.unidata.ucar.edu/software/netcdf-java/current/reference/DatasetUrls.html

### DIFF
--- a/finch/processes/utils.py
+++ b/finch/processes/utils.py
@@ -10,6 +10,8 @@ from enum import Enum
 import pandas as pd
 import xarray as xr
 import requests
+import netCDF4
+from requests.exceptions import MissingSchema, ConnectionError, InvalidSchema
 from pywps import ComplexInput, FORMATS
 from siphon.catalog import TDSCatalog
 from pywps import configuration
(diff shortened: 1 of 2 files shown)


huard:
@cehbrecht Is this something we could reuse somewhere else in birdhouse ?

huard:
@davidcaron Ready to merge ?

PR #63, opened 2020-01-16
update xarray and xclim versions

Zeitsperre:
## Overview

This PR fixes [issue id]

Changes:

* Added ...

## Additional Information


.condarc
```diff
@@ -0,0 +1,5 @@
+channel_priority: strict
+channels:
+  - birdhouse
```
tlvu: We probably do not need birdhouse channel anymore.  For which package do we need it?

PR #66, opened 2020-01-17
remove .condarc file

davidcaron:


davidcaron:
It was generated when I was trying to debug the conda build (why Gdal was being installed from the default channel and not conda-forge) and I used the `conda config --set ` command.

It's simply not needed, it was added by mistake.

https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html

tlvu:
Ah, thanks for the explanation.

PR #76, opened 2020-02-03
pin pywps==4.2.3

davidcaron:
## Overview

Make sure to freeze pywps version and test the next update before requiring it.


davidcaron:
@huard, I just wanted to pin pywps to 4.2.3. I had to disable opendap tests, because "http://test.opendap.org" was down this morning... we should either record those requests or choose another server.
Ignore the following commits, I was just fixing the merge with master.

PR #79, opened 2020-02-14
Bccaqv2 ensembles indices

davidcaron:
## Overview

Adds a list of processes that query daily BCCAQv2 datasets, subsets (grid point or bbox), computes indices and creates an ensemble.

Polygon subset will be in another PR, as the base polygon subset process is not yet implemented.

I want to draw your attention to these functions in `ensemble_utils.py`: 
 - `ensemble_utils.ensemble_common_handler`
 - `ensemble_utils.make_ensemble`

The ensemble creation is a part of the computation I'm not very familiar with, but it seems relatively straight forward.

This PR is based on #78.


davidcaron:
@tlogan2000 I've seen in your notebook examples that you save the netcdf files with a compression level of 9 and a time dtype of float32. Can you explain why you were doing so? I'm currently saving the netcdf files with compression of 0 and dtype of float32, but I'm not sure if that's the right thing to do here.

tlogan2000:
compression is up to you ... Louis David recommended 5 in the end.  The time as a float is because thredds does not seem to recognize the 'long' dtype (xarray default I suppose) and breaks the opendap link (query page) ...  I just set to float as the time vector is typically tiny in size compared ot the rest of the file

tlogan2000:
FyI i will send you an corrected / improved notebook in a few minutes...  The original had a few missing things making sure attributes are carried over duiring averaging and dropping unneeded dimensions etc etc

davidcaron:
Currently, only indicators with tasmin, tasmax and pr are available. I'm working on adding indices that require datasets that can be computed (ex: 'tas' for cold_spell_days).

davidcaron:
Sure, If my understanding is correct, the bccaqv2 processes are basically processes that work with CF-compliant files... so maybe before the next release, I could rename everything that mentions bccaqv2 with something more standard, and the input files path would be a configuration parameter.

Just a thought, I haven't looked into it too much.

tlogan2000:
I agree that some of these are a little to specific to data portal site.  Having generic processes that could point to various data sources would be best

PR #93, opened 2020-02-25
Add model input to ensemble processes

### LINKED ISSUE #91: Allow selection of models in `ensemble_*` processes
## Description

Currently, all the 24 hard-coded models are taken when calculating an ensemble. We need to add an input so that the user can choose which models to use.

davidcaron:
## Overview

This PR fixes #91

Changes:

* Added PCIC 12 models taken from https://www.pacificclimate.org/data/statistically-downscaled-climate-scenarios


finch/default.cfg
```diff
@@ -15,7 +15,8 @@ maxprocesses = 10
 parallelprocesses = 2
 
 [finch]
-bccaqv2_url = https://boreas.ouranos.ca/thredds/catalog/birdhouse/pcic/BCCAQv2/catalog.xml
+default_dataset = bccaqv2
+dataset_bccaqv2 = https://boreas.ouranos.ca/thredds/catalog/birdhouse/pcic/BCCAQv2/catalog.xml
```
tlvu: Shouldn't we use `pavics.ouranos.ca`?  `boreas` is an internal name so if Finch is deployed elsewhere than at Ouranos, it won't be accessible?  Or this is actually the intention?

davidcaron: I wasn't aware of the distinction between pavics.ouranos.ca and boreas.ouranos.ca... Both worked, so I assumed they were an alias. I'll change it to pavics when I get the chance.

tlvu: Oh, both works from CRIM side?  I am a bit surprised, maybe @moulab88 has opened a special rule in the firewall for CRIM IP range.

moulab88:
Yes both work not only for the CRIM

beluga4.int.ets1.calculquebec.ca 1000% dig pavics.ouranos.ca

;; ANSWER SECTION:
pavics.ouranos.ca. 21599 IN CNAME boreas.ouranos.ca.
boreas.ouranos.ca. 21592 IN A 207.162.87.155

tlvu:
> Yes both work not only for the CRIM

Oh sorry, my bad, I thought `boreas` only worked internally to Ouranos.

davidcaron:
@huard This branch is deployed to finch.crim.ca/wps, I tested each process type and it seems to work fine. I would be ready to merge this.

davidcaron:
Thanks for the review!

PR #95, opened 2020-03-03
Add subset shape processes

davidcaron:
## Overview

Adds polygon subsetting processes.

Only supports geojson for now... @huard, maybe we should support sending a zipped file containing a shapefile?


huard:
@davidcaron Yes, zipped shapefile is a very common scenario.

davidcaron:
Yes, it's still a WIP to chain processes.

My reasoning is that if the user is not doing the process chaining themselves, the inputs they need to provide are exactly the same... So yes, the number of processes is growing, but even if we chain the processes, there will still be the same number of processes if that makes sense. I may be missing something, but I had to do it this way for now, to avoid depending on the progression of other projects.

I'm trying to compose the operations into independent function calls, so when process chaining is possible, the migration should be less painful.

Zeitsperre:
Not sure if if helps, but within `raven`, I have a bunch of utility libraries that are used to extract shapefiles from zipped archives. The two most important are `archive_sniffer` and it dependent function, `generic_extract_archive`. I'm sure you can grab them and drop them in as utilities.

https://github.com/Ouranosinc/raven/blob/ad8f6189e6beb99b0caa4a2d153cad79e8a200d8/raven/utils.py#L78

davidcaron:
I will merge this for a `0.5.0` release, and add shapefiles later: https://github.com/bird-house/finch/issues/105

PR #97, opened 2020-03-04
Fix ensemble computation

davidcaron:
## Overview

 - add `.load()` to ensemble computation:
```
    # Depending on the datasets, I've found that writing the netcdf could hang
    # if the dataset was not loaded explicitely previously... Not sure why.
    # The datasets should be pretty small when computing the ensembles, so this is
    # a best effort at working around what looks like a bug in either xclim or xarray.
    # The xarray documentation mentions: 'this method can be necessary when working
    # with many file objects on disk.'
    ensemble.load()
```
 - Fix the broken time index in this dataset `tasmax_day_BCCAQv2+ANUSPLIN300_CESM1-CAM5_historical+rcp85_r1i1p1_19500101-21001231` when writing to disk. This ensures that it's called in all the processes. Initially, it was fixed only when tasmax and tasmin were used in heat wave computations.

Edit: added demonstration of broken time index:
```python
>>> import xarray as xr
>>> ds = xr.open_dataset('https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/pcic/BCCAQv2/tasmax_day_BCCAQv2+ANUSPLIN300_CESM1-CAM5_historical+rcp85_r1i1p1_19500101-21001231.nc')
>>> ds.time[31688:31693].values
array([cftime.DatetimeNoLeap(2036-10-26 12:00:00), 
       cftime.DatetimeNoLeap(2036-10-27 12:00:00),
       cftime.DatetimeNoLeap(1850-01-01 00:00:00),
       cftime.DatetimeNoLeap(2036-10-29 12:00:00),
       cftime.DatetimeNoLeap(2036-10-30 12:00:00)], dtype=object)
>>> ds.time[31690].values 
array(cftime.DatetimeNoLeap(1850-01-01 00:00:00), dtype=object)
```

### DIFF
--- a/finch/processes/ensemble_utils.py
+++ b/finch/processes/ensemble_utils.py
@@ -441,6 +441,13 @@ def make_file_groups(files_list: List[Path]) -> List[Dict[str, Path]]:
 
 def make_ensemble(files: List[Path], percentiles: List[int]) -> None:
     ensemble = ensembles.create_ensemble(files)
+    # Depending on the datasets, I've found that writing the netcdf could hang
+    # if the dataset was not loaded explicitely previously... Not sure why.
+    # The datasets should be pretty small when computing the ensembles, so this is
+    # a best effort at working around what looks like a bug in either xclim or xarray.
+    # The xarray documentation mentions: 'this method can be necessary when working
+    # with many file objects on disk.'
+    ensemble.load()
     # make sure we have data starting in 1950
     ensemble = ensemble.sel(time=(ensemble.time.dt.year >= 1950))
     ensemble_percentiles = ensembles.ensemble_percentiles(ensemble, values=percentiles)
--- a/finch/processes/ensemble_utils.py
+++ b/finch/processes/ensemble_utils.py
@@ -405,16 +405,16 @@ def make_file_groups(files_list: List[Path]) -> List[Dict[str, Path]]:
 
 def make_ensemble(files: List[Path], percentiles: List[int]) -> None:
     ensemble = ensembles.create_ensemble(files)
+    # make sure we have data starting in 1950
+    ensemble = ensemble.sel(time=(ensemble.time.dt.year >= 1950))
+    ensemble_percentiles = ensembles.ensemble_percentiles(ensemble, values=percentiles)
     # Depending on the datasets, I've found that writing the netcdf could hang
     # if the dataset was not loaded explicitely previously... Not sure why.
     # The datasets should be pretty small when computing the ensembles, so this is
     # a best effort at working around what looks like a bug in either xclim or xarray.
     # The xarray documentation mentions: 'this method can be necessary when working
     # with many file objects on disk.'
-    ensemble.load()
-    # make sure we have data starting in 1950
-    ensemble = ensemble.sel(time=(ensemble.time.dt.year >= 1950))
-    ensemble_percentiles = ensembles.ensemble_percentiles(ensemble, values=percentiles)
+    ensemble_percentiles.load()
 
     return ensemble_percentiles
 
--- a/finch/processes/ensemble_utils.py
+++ b/finch/processes/ensemble_utils.py
@@ -13,7 +13,6 @@ from siphon.catalog import TDSCatalog
 import xarray as xr
 import xclim
 from xclim import ensembles
-from xclim.checks import assert_daily
 from xclim.utils import Indicator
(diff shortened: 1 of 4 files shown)


huard:
@tlogan2000 I'll let you review this one.

finch/processes/ensemble_utils.py
```diff
@@ -441,6 +405,13 @@ def make_file_groups(files_list: List[Path]) -> List[Dict[str, Path]]:
 
 def make_ensemble(files: List[Path], percentiles: List[int]) -> None:
     ensemble = ensembles.create_ensemble(files)
+    # Depending on the datasets, I've found that writing the netcdf could hang
+    # if the dataset was not loaded explicitely previously... Not sure why.
+    # The datasets should be pretty small when computing the ensembles, so this is
+    # a best effort at working around what looks like a bug in either xclim or xarray.
+    # The xarray documentation mentions: 'this method can be necessary when working
+    # with many file objects on disk.'
+    ensemble.load()
```
tlogan2000: I guess this is ok if subsets are relatively small but would likely 1) take a long time 2) use large amts of memory  if regions become very big... 
Would a .load() on or after line 417 work as well?
i.e.
```
ensemble_percentiles = ensembles.ensemble_percentiles(ensemble, values=percentiles).load()

```

finch/processes/utils.py
```diff
@@ -507,15 +509,49 @@ def make_tasmin_tasmax_pairs(
         )
 
 
+def fix_broken_time_index(ds: xr.Dataset):
```
tlogan2000: What is this doing?  Does an .nc file have a 1850-1-1 datestamp?  I might have missed this as in most of my scripts I select time>=1950  before doing anything

tlogan2000: Should we just fix this on disk? ... i.e. rewrite that file with a corrected timestamp instead of having a 'correction' function in finch?

finch/processes/ensemble_utils.py
tlogan2000: This .load() should use less memory as we are collapsing the 24 models down to 3 variables

tlogan2000: Could we also maybe try to use 'persist'? 
http://xarray.pydata.org/en/stable/generated/xarray.Dataset.persist.html

davidcaron: Good observations, I will look into this.

finch/processes/utils.py
tlogan2000: Just read the description of the PR re: the problematic file ... sorry

davidcaron: I agree, it should probably be fixed on disk...

``` python
>>> import xarray as xr
>>> ds = xr.open_dataset('https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/pcic/BCCAQv2/tasmax_day_BCCAQv2+ANUSPLIN300_CESM1-CAM5_historical+rcp85_r1i1p1_19500101-21001231.nc')
>>> ds.time[31688:31693].values
array([cftime.DatetimeNoLeap(2036-10-26 12:00:00), 
       cftime.DatetimeNoLeap(2036-10-27 12:00:00),
       cftime.DatetimeNoLeap(1850-01-01 00:00:00),
       cftime.DatetimeNoLeap(2036-10-29 12:00:00),
       cftime.DatetimeNoLeap(2036-10-30 12:00:00)], dtype=object)
>>> ds.time[31690].values 
array(cftime.DatetimeNoLeap(1850-01-01 00:00:00), dtype=object)
```

Here I'm trying to fix it on the fly, with as less overhead as possible.

tlogan2000: I guess leave it as is for now... we can fix the file when we have time then remove or not this subfunction as desired

finch/processes/ensemble_utils.py
davidcaron: I tried using `persist` and it doesn't work. Looking at the code, it seems to only change datasets containing dask dataframes.

Delaying the call to `.load()` like you mentioned worked though... so it should use less memory. 

For what it's worth, I experienced this process hang only with rcp26 ensembles with the 24 models.

tlogan2000: Ok sounds good. I'm ok with the changes so feel free to merge

PR #52, opened 2019-11-19
added make command to run pytest nbval

huard:
## Overview

This PR fixes #49

Changes:

* Added make command to test notebooks.

PR #112, opened 2020-03-16
run notebooks in travis ci

### LINKED ISSUE #49: Test notebooks in Travis-CI
## Description

Similar to what is done in Raven.

davidcaron:
## Overview

This PR fixes #49


davidcaron:
🤦‍♂ @huard I just duplicated https://github.com/bird-house/finch/pull/52

docs/source/notebooks/finch-usage.ipynb
```diff
@@ -18,8 +18,8 @@
     "import os\n",
     "from birdy import WPSClient\n",
     "#url = 'https://finch.crim.ca/wps'\n",
-    "url = 'https://pavics.ouranos.ca/twitcher/ows/proxy/finch/wps'\n",
-    "#url = 'http://localhost:5000'\n",
+    "#url = 'https://pavics.ouranos.ca/twitcher/ows/proxy/finch/wps'\n",
+    "url = 'http://localhost:5000'\n",
```
tlvu: Can not hardcode `localhost` like that.  Have to leave `pavics.ouranos.ca/twitcher/...` to not break Jenkins later (Jenkins runs the exact same notebook).

Get the wps url like so `url = os.environ.get("RAVEN_WPS_URL", "https://pavics.ouranos.ca/twitcher/ows/proxy/raven/wps")` (https://github.com/Ouranosinc/raven/blob/master/docs/source/notebooks/Delineation_workflow.ipynb) and in Travis override it to the localhost.  Probably Raven also have the Travis override example.

requirements_dev.txt
```diff
@@ -4,3 +4,5 @@ pytest-flake8
 sphinx>=1.7
 bump2version
 twine
+nbval
+birdy
```
tlvu: on Pypi it's `birdhouse-birdy` !

huard:
It was messy anyway. But one thing that is useful is to do like in Raven, and have an environment variable for the local WPS host url. This lets us run the notebooks on the local or remote server without modifying the notebook. 
So in the notebook:
```
url = os.environ.get("FINCH_WPS_URL", "https://pavics.ouranos.ca/twitcher/ows/proxy/finch/wps")
```
and in the makefile something like:
```
.PHONY: test_nb
test_nb:
	@echo "Running notebook-based tests"
	@bash -c "source $(ANACONDA_HOME)/bin/activate $(CONDA_ENV);env FINCH_WPS_URL=$(FINCH_WPS_URL) pytest --nbval $(CURDIR)/docs/source/notebooks/ --sanitize-with $(CURDIR)/docs/source/output_sanitize.cfg --ignore $(CURDIR)/docs/source/notebooks/.ipynb_checkpoints"
```

docs/source/notebooks/finch-usage.ipynb
```diff
@@ -81,13 +81,15 @@
      "text": [
       "ProcessSucceeded\n",
       "frost_daysResponse(\n",
-      "    output_netcdf='https://pavics.ouranos.ca/wpsoutputs/e8afbb04-42d3-11ea-9531-0242ac12000b/out.nc',\n",
-      "    output_log='https://pavics.ouranos.ca/wpsoutputs/e8afbb04-42d3-11ea-9531-0242ac12000b/log.txt'\n",
+      "    output_netcdf='http://localhost:5000/outputs/50c0a3f8-67c7-11ea-9e2d-b06ebf31cced/frost-days_SRES-A2-experiment_20460101-20650101.nc',\n",
```
tlvu: Should also keep `https://pavics.ouranos.ca/wpsoutputs` to not break Jenkins later.

huard: Ideally those differences would be catched by the `output_sanitize.cfg`, so the tests pass both on Jenkins and Travis-CI.

tlvu: > Ideally those differences would be catched by the `output_sanitize.cfg`, so the tests pass both on Jenkins and Travis-CI.

Agreed, I'll look into this next time I am around notebook tests.

requirements_dev.txt
davidcaron: Thanks, done

docs/source/notebooks/finch-usage.ipynb
davidcaron: done

docs/source/notebooks/finch-usage.ipynb
davidcaron: This output is actually skipped by a `NBVAL_IGNORE_OUTPUT` comment, so tests will pass either way on jeknins. Which is good, because uuid will never correspond, unless we do some `output_sanitize.cfg` regex.

tlvu: Oh !  Did not realize there is a `NBVAL_IGNORE_OUTPUT`!  Wonder why it was added since we do have regex to sanitize the uuid.

davidcaron: Could you paste the regex that skips uuids? It doesn't seem to be in this repo?

tlvu: This repo should curl it from https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/raw/master/notebooks/output-sanitize.cfg so there's only 1 copy of this file.  Raven should also doing this, I hope !  Must have sample code from Raven.

I am actually trying to add regex for this output url here https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/raw/add-regex-to-output-sanitize-cfg-for-Travis-CI/notebooks/output-sanitize.cfg can you help test so we can remove the `NBVAL_IGNORE_OUTPUT`?

tlvu: Oh wait, the `NBVAL_IGNORE_OUTPUT` was for the `print(log)` cell, not the `print(out)` cell that shows the wps output url.

davidcaron: What about adding the output-sanitize.cfg to the bird-house cookie cutter? It seems a bit overkill to me to add this curl command for a single cell... and the fact that the cell executed without errors is a good indication that there was no error... I don't know, I'll let you decide if it's worth the effort.

tlvu: We do not refresh from the cookie-cutter fast/frequent enough, so I'd rather do not have the `output-sanitize.cfg` there.

The cell execute without error in Travis-CI, but will not in Jenkins.

What we can do is add the curl command to travis-ci file in the cookie-cutter ! ... if the cookie-cutter has such a file and we are good for all the future notebooks for all the future birds.

Oh wait, the entire notebook passed on Travis-CI even without the regex for the UUID?  What?!!!

tlvu: Looked at Travis-CI logs
```
$ make test-notebooks

make: Nothing to be done for 'test-notebooks'.

The command "make test-notebooks" exited with 0.
```

The notebooks test did not run, so of course it passed !

docs/source/notebooks/finch-usage.ipynb
```diff
@@ -81,13 +81,15 @@
...
-      "    output_log='https://pavics.ouranos.ca/wpsoutputs/e8afbb04-42d3-11ea-9531-0242ac12000b/log.txt'\n",
+      "    output_netcdf='http://https://pavics.ouranos.ca/wpsoutputs/e8afbb04-42d3-11ea-9531-0242ac12000b/frost-days_SRES-A2-experiment_20460101-20650101.nc',\n",
+      "    output_log='https://pavics.ouranos.ca/wpsoutputs/e8afbb04-42d3-11ea-9531-0242ac12000b/log.txt',\n",
+      "    ref='https://pavics.ouranos.ca/wpsoutputs/e8afbb04-42d3-11ea-9531-0242ac12000b/input.meta4'\n",
       ")\n"
      ]
     }
    ],
    "source": [
+    "# NBVAL_IGNORE_OUTPUT\n",
```
tlvu: Oh you added `NBVAL_IGNORE_OUTPUT` !  Sorry I didn't saw this one, can you move it?

docs/source/notebooks/finch-usage.ipynb
tlvu: My bad, Raven has the `output-sanitize.cfg` physically committed in the repo https://github.com/Ouranosinc/raven/blob/a4dc8ebb85a7c239596852929509f6aa5b5fa245/docs/source/output_sanitize.cfg and I approved that gigantic PR it was part of.  Brown paper bag on me.

Can you just do it proper this time so there is no duplicate of that file and I'll tell @Zeitsperre to fix on Raven side?

Sorry for not catching earlier on Raven side so you would already have sample code.

davidcaron:
@tlvu please check https://github.com/bird-house/finch/pull/112/commits/fadbedb118bd57badd7bb909a2b2d424c85d8d4e
It seems `pavics.ouranos.ca/wpsoutputs` doesn't get replaced to `localhost:5000/outputs` by the output_sanitize.cfg file

tlvu:
@davidcaron @huard Is Finch docker image 0.4.1 good enough or we need an new Finch docker release?  Jenkins will be broken with this notebook update.

The notebook is also auto-deployed hourly to Jupyter tutorial-notebooks folder so that demo notebook is also broken for users as well.

davidcaron:
> Jenkins will be broken with this notebook update.

Why does Jenkins test the master branch, and not the version of finch that is deployed on the tested server?

> The notebook is also auto-deployed hourly to Jupyter tutorial-notebooks folder so that demo notebook is also broken for users as well.

Same question, why are the notebooks taken from the master branch, and not the release version? Am I missing something, or it makes updating them much more complicated?

tlvu:
> > Jenkins will be broken with this notebook update.
> 
> Why does Jenkins test the master branch, and not the version of finch that is deployed on the tested server?

Jenkins was not meant to replace Travis-CI.  Jenkins test against the production server to ensure real user experience is not broken while Travis-CI tests against the development version for the developer experience.

> > The notebook is also auto-deployed hourly to Jupyter tutorial-notebooks folder so that demo notebook is also broken for users as well.
> 
> Same question, why are the notebooks taken from the master branch, and not the release version? Am I missing something, or it makes updating them much more complicated?

If taken from the release version, any notebooks improvements on master that do not require a change to the deployed server can not be deployed.

Agreed it puts a pressure on releasing the server if the notebook changes need new server.

I do not have a perfect solution unfortunately.  But I think it is more beneficial to rather be on the frequent/fast release of notebooks/server than the reverse.

It's okay if Jenkins is broken for a day or two.  Maybe a simpler solution next time is to hold off releasing not backward compatible notebooks to master or to write backward compatible notebooks.

tlvu:
@davidcaron thanks for the question, that made me think about fixing the notebook code so it is backward compatible with old server, see commit https://github.com/bird-house/finch/pull/114/commits/1608df81a7d1df2809ffc57e16cbe770bbe20cf9

PR #119, opened 2020-03-24
Remove realization csv

davidcaron:
## Overview

@tlogan2000 Does it need to be removed from the netcdf output also?


tlogan2000:
> ## Overview
> @tlogan2000 Does it need to be removed from the netcdf output also?

I think I would simply get rid of if in the netcdf 'percentiles' dataset. It's less problematic in netcdf as it is a simply a dimension that isn't 'used' but might as well get rid of it.  Will solve the issue in the output csv as well as it is no longer even there

Probably directly in make_ensemble (added line in the middle should do the trick) :
```

def make_ensemble(files: List[Path], percentiles: List[int]) -> None:
    ensemble = ensembles.create_ensemble(files)
    # make sure we have data starting in 1950
    ensemble = ensemble.sel(time=(ensemble.time.dt.year >= 1950))
    ensemble_percentiles = ensembles.ensemble_percentiles(ensemble, values=percentiles)
    
    # added line 
    ensemble_percentiles = ensemble_percentiles.drop('realization')
    


    # Depending on the datasets, I've found that writing the netcdf could hang
    # if the dataset was not loaded explicitely previously... Not sure why.
    # The datasets should be pretty small when computing the ensembles, so this is
    # a best effort at working around what looks like a bug in either xclim or xarray.
    # The xarray documentation mentions: 'this method can be necessary when working
    # with many file objects on disk.'
    ensemble_percentiles.load()

    return ensemble_percentiles

**Comment by tlogan2000** [comment]
This might be something we implement directly in xclim (i.e. clean up the output from ```ensembles.ensemble_percentiles()```) but might not have time to implement the change / push a new version quickly enough for the analysis page stuff

**Comment by davidcaron** [comment]
@tlogan2000 please take a look at the latest commit

**Comment by davidcaron** [comment]
@Zeitsperre any idea why the Travis-ci hook is broken?

**Comment by davidcaron** [comment]
Or maybe @Zeitsperre doesn't have the permissions here, @huard any idea?

**Comment by tlogan2000** [comment]
> Or maybe @Zeitsperre doesn't have the permissions here, @huard any idea?

@tlvu do you have an idea why the travis build is not triggering here?   Are you the repo admin?

**Comment by davidcaron** [comment]
It seems like travis is building the commits correctly, but github is not recieving the status https://travis-ci.org/github/bird-house/finch/builds/666970536?utm_medium=notification&utm_source=github_status

**Comment by davidcaron** [comment]
Oh, and the `dap_subset` test is flaky, needs to be restarted, but I seem to not have the permissions anymore ☹️

**Comment by tlvu** [comment]
I am not Travis-CI admin, ping @Zeitsperre

**Comment by davidcaron** [comment]
It seems to be a Travis-CI issue, and they mention it's resolved: https://www.traviscistatus.com/incidents/rx6fhs3wqcln?u=49rdxf5zfpg9

But my last commit didn't work 😢

**Comment by Zeitsperre** [comment]
Sorry to not see this. Yes, this is a Travis CI issue. I'll see if I can find out more on the status now. You may need to push another commit to trigger it when they claim the issue is fixed.

**Comment by MatProv** [comment]
@Zeitsperre It was due to xarray 0.15.1 (released two weeks ago) now displaying notebook cell outputs as html by default, not plain text.

**Comment by tlvu** [comment]
@davidcaron @MatProv you guys bump the version in this PR, please push a new tag as well, because right now there is no docker build for this new version.

Jenkins has been failing over the week-end with the following error and I think we just need a new Finch to match the notebook change
```
20:35:52  _________ finch-master/docs/source/notebooks/dap_subset.ipynb::Cell 9 __________
20:35:52  Notebook cell execution failed
20:35:52  Cell 9: Cell outputs differ
20:35:52  
20:35:52  Input:
20:35:52  resp = wps.sdii(pr + sub)
20:35:52  out = resp.get(asobj=True)
20:35:52  out.output_netcdf.sdii
20:35:52  
20:35:52  Traceback:
20:35:52   mismatch 'text/html'
20:35:52  
20:35:52   assert reference_output == test_output failed:
20:35:52  
20:35:52    '<pre>&lt;xar...vera...</pre>' == '<pre>&lt;xar...vera...</pre>'
20:35:52    Skipping 350 identical leading characters in diff, use -v to show
20:35:52      m/day
20:35:52    -     cell_methods:   time: mean (interval: 30 minutes)
20:35:52          history:        pr=max(0,pr) applied to raw data;\n[DATE_TIME] ...
20:35:52    +     cell_methods:   time: mean (interval: 30 minutes) 
20:35:52          standard_name:  lwe_thickness_of_precipitation_amount
20:35:52          long_name:      Average precipitation during wet days (sdii)
20:35:52          description:    Annual simple daily intensity index (sdii) : annual avera...</pre>
```

MatProv:
@tlvu Can you please retry to run the Jenkins pipeline with the 0.5.2 tag that @huard pushed yesterday?

tlvu:
@MatProv the 0.5.2 tag from huard worked, thanks for follow up.

PR #133, opened 2020-11-26
Fix 131

### LINKED ISSUE #127: Support subset with multiple gridpoints
## Description

Upgrade to xclim 0.16

### LINKED ISSUE #131: Update to xclim 0.21
## Description

- [x] Make use of indicator registry
- [x] Add new indices
- [x] Relies on clisops for subsetting
- [x] Add option for missing value handling

huard:
## Overview

This PR fixes #131 
Fixes #127 

Changes:

* Pins xclim==0.21
* Work with Indicator registry
* Depends on clisops (0.4.0)
* Adds support for missing value handling options. 

## Related Issue / Discussion
Regression in `test_wps_xsubset_polygon`. Point at 0,0 is not included in subset, as it was in previous versions. 

## Additional Information


finch/processes/ensemble_utils.py
```diff
@@ -343,7 +343,7 @@ def make_output_filename(process: Process, inputs: List[PywpsInput]):
 def uses_accepted_netcdf_variables(indicator: Indicator) -> bool:
     """Returns True if this indicator uses  netcdf variables in `accepted_variables`."""
 
-    params = eval(indicator.json()["parameters"])
```
Zeitsperre, line 346: Good catch.

Zeitsperre (approved):
I can't see any problems, but I'm not very familiar with Finch. I'd consider asking for another person's review before merging.

huard:
@aulemahal Starting point for xclim update. 

Related issue found while trying to fix subset_shape regression: https://github.com/roocs/clisops/issues/95

Zeitsperre:
Just tried with updated clisops. `test_wps_subsetpoly` is failing for good reasons here. Without `buffer` being added, `lat` at location `[0]` is only being touched. Enlarging the zone to include `0`, by way of `buffer` or by changing the polygon coordinates, the proper output is received.

huard:
I think this is a surprising behavior. In my mind, the boundary is inside the shape.

tlogan2000:
FYI for the CCDP project we will actually need finch to include this xclim PR that fixes the degree day exceedance date to work with dask: 
https://github.com/Ouranosinc/xclim/pull/612

@aulemahal @huard @Zeitsperre : Opinions on whether we should do a patch release?

Zeitsperre:
> @aulemahal @huard @Zeitsperre : Opinions on whether we should do a patch release?

Our next minor release is due out in two weeks. We could try adapting this PR and #131 to make use of it? What determines a patch release is the need for one. Is this high priority? If not, we can bring xclim @ v0.22 to this PR instead and perform another PR once 0.23 is out. Just my thoughts

tlogan2000:
For timeline : There's a CCDP (climatedata.ca) deliverable for end of march to integrate the DDED calculation into the analysis page (i.e. on the fly user calculations) ... which runs finch processes under the hood.  Our next step is to actaully get the DDED algorithm into the finch server but will then need to coordinate with @MatProv from CRIM to get it running correctly for the site etc.   So basically there is some flexibility on time but not all that much in my opinon   

If a finch update is time consuming then maybe better to do it a single time?  On the other hand it would be good to get better at regularly updating / syncing  xclim and finch releases

tlogan2000:
> Just tried with updated clisops. `test_wps_subsetpoly` is failing for good reasons here. Without `buffer` being added, `lat` at location `[0]` is only being touched. Enlarging the zone to include `0`, by way of `buffer` or by changing the polygon coordinates, the proper output is received.

Not sure I follow all the details here but if the the polygon exactly touches a grid centroid I would consider this an "intersection" (in GIS terms) and the subset should include the gridpoint..  Possible to look at a 'diff' in the clisops code to see what has changed?

Zeitsperre:
I'm currently reading over the changes, but aside from using `vectorize` version of `create_mask`, I don't see any major alterations. My guess is that the issue is tucked away in there.

huard:
This will be ready to be merged when clisops makes a new release.

Zeitsperre:
@huard This PR looks good to me. Merge when you're satisfied with it.

finch/processes/__init__.py
```diff
@@ -25,22 +24,36 @@
...
+    from xclim.core.indicator import registry
+
+    out = {}
+    for key, val in module.__dict__.items():
+        if hasattr(val, "identifier"):
+            name = val.identifier.upper()
+            if name in registry:
+                out[key] = val
+
+    return OrderedDict(sorted(out.items())).values()
```
aulemahal, line 41: Is there a reason we are restricting the parsing to a module? With 0.21, we have  `.get_instance()` and  `.realm`. Doing:
```
indicators = [ind.get_instance() for name,  ind in registry.items() if ind.realm == 'atmos' and name not in not_implemented]
```
would simplify the process, and would open the door for externally-declared indicators, which was what the registry was implemented for!

aulemahal (approved):

finch/processes/__init__.py
huard: Indeed. Fixed.

docs/source/notebooks/finch-usage.ipynb
```diff
@@ -47,15 +47,19 @@
      "text": [
       "Help on method frost_days in module birdy.client.base:\n",
       "\n",
-      "frost_days(tasmin=None, freq='YS') method of birdy.client.base.WPSClient instance\n",
+      "frost_days(tasmin=None, missing_options=None, check_missing='any', freq='YS') method of birdy.client.base.WPSClient instance\n",
```
tlvu, line 50: This change breaks Jenkins (against Finch deployed on PAVICS).  Can we have a new release of Finch to deploy to PAVICS?

huard: Yes, see https://github.com/bird-house/finch/pull/138

PR #135, opened 2020-12-21
Subset-shape with zipped shapefiles

### LINKED ISSUE #105: Accept shapefile input to subset shape processes
## Description

Allow a zipped shapefile input to subset shape processes.

Zeitsperre:
## Overview

This PR fixes #105 

Changes:

* Added `FORMATS.SHP` to the `wpsio.py` as an accepted complex input
* Added a small zipped Shapefile for local testing
* Added a function to expose this shapefile (not sure if this is appropriate/follows conventions)

## Related Issue / Discussion

~This is currently not working.~ **Seems to be working now.**

There seems to be some complexity with allowing zipped shapefiles as it relies on calls to GDAL to handle IO processes. Seeing as I am very unfamiliar with the way finch handles IO, I need guidance in how best to store the bytes as a temporary file or how best to read the zip file bytes directly to subset. Approaches I have tried thus far:

- Reading the data into a temporary `zip` file and sending that to `clisops.core.subset` (This isn't handled properly; we've never needed this feature, but it could be done --> requires changes in clisops)
- **Sending the zipped dataset directly to [`Geopandas.read_file()`](https://geopandas.org/io.html) using the `!`path to the `POLYGON.shp` (This isn't understoud in the `execute_process()` call)** - **_It looks like this was the best option to move forward._**
- Receiving the file in `finch/processes/subset.py`, writing the file into a [`Fiona.ZipMemoryFile`](https://fiona.readthedocs.io/en/latest/fiona.html?highlight=zip#fiona.io.ZipMemoryFile) and sending that to `clisops.core.subset` (This fails early due to something about how the Bytes received cannot be understoud).

Zipped shapefiles are really clunky to deal with as there needs to be many helper processes if we want to simply it. Some functions similar to those present in RAVEN (I've named them `sniffer` functions) are needed in order to gather information from the zip files (ie: layer names, CRSes) could be ported/adapted for this code, but I require information on how best to receive and store the bytes in memory or /tmp locations. 

## Additional Information

https://github.com/Toblerity/Fiona/issues/753
https://github.com/Toblerity/Fiona/issues/318
https://fiona.readthedocs.io/en/latest/fiona.html?highlight=zip#fiona.io.ZipMemoryFile


huard:
Write a test for this, and I'll take a look at your branch.

huard:
Look at my commit, should fix your issue.

Zeitsperre:
@huard This is passing for me locally. Should be good to go.

huard (approved):

PR #139, opened 2021-01-22
Add downscaling processes

### LINKED ISSUE #126: Process for downscaling
Once xclim offers downscaling methods, create WPS wrapper.

huard:
## Overview

This PR fixes #126 

Changes:

* Added empirical_quantile_mapping process

## Additional Information

We need to see how to generalize to other sdba methods. 


finch/processes/wps_sdba.py
```diff
@@ -0,0 +1,220 @@
...
+                else:
+                    train[key] = single_input_or_none(request.inputs, key)
+
+            _log("Successfully read inputs from request.", 1)
+
+            group = xclim.sdba.Grouper(**group)
+            _log("Grouper object created.", 2)
+
+            bc = xclim.sdba.EmpiricalQuantileMapping(**train, group=group)
+            bc.train(res["ref"], res["hist"])
```
aulemahal, line 184: I suggest we convert ref, hist and sim to a no_leap calendar by default before running the adjustment. This could also be controlled by an argument, but it doesn't make much sense to not do it anyway.

finch/processes/wps_sdba.py
```diff
@@ -0,0 +1,220 @@
...
+            _log("Training object created.", 3)
+
+            out = bc.adjust(res["sim"], interp=group["interp"]).to_dataset(name=name)
```
aulemahal, line 187: I think you forgot to pass the adjustment kwargs (extrapolation).

finch/processes/wps_sdba.py
```diff
@@ -0,0 +1,220 @@
...
+                dt=1,
+            ):
+                dataset_to_netcdf(out, out_fn)
+
+        except Exception as exc:
+            LOGGER.exception(exc)
+            err_msg = traceback.format_exc()
+            raise ProcessError(
+                err_msg, max_length=len(err_msg), allowed_chars=string.printable
+            ) from exc
```
aulemahal, line 205: This is already done in `FinchProcess._handler_wrapper`. But here you have a bit stuff you could move there.

finch/processes/wps_sdba.py
```diff
@@ -0,0 +1,220 @@
...
+                ),
+                LiteralInput(
+                    "kind",
+                    "Kind of adjustment (+, *)",
+                    abstract="Use * for multiplicative adjustment, or + for additive adjustement.",
+                    data_type="string",
+                    default="+",
+                    allowed_values=["+", "*"],
+                    min_occurs=0,
+                ),
```
aulemahal, line 134: If this is too much trouble, it could also be a boolean? I don't think we'll implement other "kinds", as the _new way_ will arrive first.

aulemahal (approved):
Well, clean and simple!

finch/processes/wps_sdba.py
huard: ok. Please see commit to confirm I got it right.

finch/processes/wps_sdba.py
huard: youpelai comme tu dis.

finch/processes/wps_sdba.py
huard: Complicates testing though...

huard:
New xarray release broke the build. 
http://xarray.pydata.org/en/stable/whats-new.html

aulemahal:
Fixed the error in roocs/clisops#169. We will be able to unpin on the next release of clisops.

PR #147, opened 2021-01-26
Fix 146 - Release 0.6.1

### LINKED ISSUE #146: Release 0.6.1
## Description

Forgot to bumpversion for the 0.6.0 release. 
Complete release notes.

huard:
## Overview

This PR fixes #146

Changes:

* Added change notes
* Fixed bumpversion config
* Bumped version to 0.6.1


setup.cfg
```diff
@@ -11,7 +11,7 @@ search = __version__ = '{current_version}'
 replace = __version__ = '{new_version}'
 
 [bumpversion:file:docs/source/conf.py]
-search = version|release = {current_version}
+parse = release|version = "{current_version}"
```
tlvu, line 14: oh !  This `search` to `parse` needs to be backported to the cookiecutter I think.

tests/test_client.py
```diff
@@ -0,0 +1,26 @@
+from pywps import Service
+from .common import client_for
+from finch.processes import get_processes
+import pytest
+
+
+class TestClient:
+    URL = "http://localhost:9099"
```
tlvu, line 8: Very minor suggestion, maybe allow override using environment var. So something like `URL = os.environ.get('WPS_URL', "http://localhost:9099")` which would allow to use the same test client to maybe target other Finch, ex the Finch in prod.

tlvu (approved):
LGTM, minor suggestion, for another PR if you wish.

setup.cfg
huard: Yes, I opened an issue.

tests/test_client.py
huard: Done.

huard: In fact this is a dummy value. It's not hitting the server running on this port, I'm creating a Service and getting the xml documents, then using this to instantiate client objects. I've changed the URL to `URL = "http://dummy_url.org"` to clarify this.

PR #152, opened 2021-02-26
Average over shape process

### LINKED ISSUE #98: Average over shape process
## Description

Given one or multiple polygons and a netCDF file, compute the spatial average (area-weighted) over each region and store along a new "geometry" dimension.

Might require to pass a file storing the cell area for accurate computations.

aulemahal:
## Overview

This PR fixes #98

Changes:

* Added a `AveragePolygonProcess` (with `subset.finch_average_shape` and `wpsio.tolerance`).

## Related Issue / Discussion
Since xESMF's `SpatialAverager` is now wrapped into clisops, we can make it available here!

## Additional Information
While testing this I stumbled upon an issue with high-resolution polygons. Well more a performance caveat : xESMF "exact" weight masks can take a long time to build when the polygons have too many nodes. I introduced here a `tolerance` kwarg to simplify the polygons using shapely's algos (through geopandas). Maybe I should move that cde to clisops? Moreover, having it in clisops could allow specifying it as a percent of the gridsize (instead of degrees like now).

Secundo, I kept the time subsetting code to imitate all the other subset processes. Would be easy to remove if not needed.

Tests are still missing.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -53,7 +53,7 @@ extensions = [
 # inherit without having to test which work which do not.
 if os.environ.get('READTHEDOCS') == 'True':
     autodoc_mock_imports = ["numpy", "xarray", "fiona", "rasterio", "shapely",
-                            "osgeo", "pandas", "statsmodels",
+                            "osgeo", "pandas", "geopandas", "statsmodels",
                             "affine", "rasterstats", "spotpy", "matplotlib",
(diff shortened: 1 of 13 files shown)


huard (approved):
Add a smoke test (just calling the process and retrieving the outputs) and we're good to go.

aulemahal:
Voici qui est fait.
I'm still taking a note that the "tolerance" trick should be in clisops, someday.

PR #164, opened 2021-03-25
Release 0.7.1

aulemahal:
## Overview

This PR makes release 0.7.1. Also added updates to CHANGES.rst and AUTHORS.rst, we forgot to update those in the last PRs.
Ran bumpversion.


huard (approved):

tlvu:
Curious, when was the tag `v0.7.1` pushed?  There is no DockerHub build for it !

tlvu:
Just manually triggered a DockerHub build for `v0.7.1`.

aulemahal:
Sorry, it's my first time doing a release! I pushed the tag with the last commit _of this PR_. I think I should have left github create the tag at the same time as the release, is that it?
Also, should I tag you when we push changes that "break" the notebooks? Or is this automated dockerhub build enough?

tlvu:
> Sorry, it's my first time doing a release! I pushed the tag with the last commit _of this PR_. I think I should have left github create the tag at the same time as the release, is that it?

You did everything right.  Github do not create the tag.  It should be created and pushed with the release PR.  So looks like the DockerHub build trigger somehow did not work for that tag.  Weird nothing changed for that automatic trigger that was working before.  Let's blame network issues or the moon.

> Also, should I tag you when we push changes that "break" the notebooks? Or is this automated dockerhub build enough?

Yes, let me know when you make a Finch release.  Unfortunately the update of Finch into the PAVICS stack is still manual.

Everything is fine, don't worry.  I am just the kind that like to find root causes when there's anomalies.

huard:
I think Github creates the tag when we use the "Release" page (if it does not already exist).

tlvu:
> I think Github creates the tag when we use the "Release" page (if it does not already exist).

Oh I didn't know this "feature".

But I think we should just stick to the known working release procedure here https://github.com/bird-house/finch/blob/master/docs/source/dev_guide.rst#bump-a-new-version

I think this is what Pascal followed anyways so I think the glitch in this case is either Github not firing the trigger to DockerHub or DockerHub somehow did not receive or did not act on the trigger.

PR #167, opened 2021-03-30
Improve logging and workaround clisops logging issue

### LINKED ISSUE #129: Improve error reporting from processes
## Description

It would be useful for debugging to get more informative messages than "process failed". 
Can we pass a traceback where potentially sensitive information would be hidden ?

### LINKED ISSUE #166: Log file not written to
## Description

Whether it is with the `cli` access or the wsgi app (with gunicorn), the log file we specify in the configuration is not ~created nor~ written to. With a vanilla `finch start`, there should be a `finch.log` file where the log would be dumped.

`emu` has the exact same config template and `cli.py` file, but there, the log file is used.

EDIT: The file is created if it doesn't exist, but the warnings and the internal error tracebacks are going into the console, **not** in the file. Is that what we want?

aulemahal:
## Overview

This PR fixes #166 and fixes #129

Changes:

* Added a workaround for the behaviour of clisops that was disabling the pyWPS log.
* Do not log warnings for the `ds` argument of xclim indicator that is not used by finch anyway,
* But do pass the identifier of the indicator, so that when unknown inputs are encountered the logged warning is more explicit.
* Catch _any_ error that happens when handling a process and re-emit as a pyWPS ProcessError, so that at least the error text is transmitted to the client.

## Additional Information
The doc of pyWPS says that we should get the traceback in the ProcessError when we use the "DEBUG" logging level. It was not able to do that, but at least the error text is transmitted.

I had not read the thread of #129 when I wrote this code. My understanding is that this solution is too broad? insecure?

As for debugging tracebacks, they are readable in the log, server side, even with the try-except.


finch/default.cfg
```diff
@@ -23,4 +23,5 @@ subset_threads = 1
 [logging]
 level = INFO
 file = finch.log
-format = %(asctime)s] [%(levelname)s] line=%(lineno)s module=%(module)s %(message)s
+database = sqlite:///pywps-logs.sqlite
```
tlvu, line 26: FYI, Finch WPS config in prod: https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/config/finch/wps.cfg.template.  Maybe (not sure on this one) this `default.cfg` is used when a config is not specified in the override so be careful when you change config here.

In this case the `database` change is probably fine since we override with Postgres in prod.

aulemahal: This change is not necessary anyway. 
Finch's wsgi app is created with this config file at the top and other passed config files are appended to the list.
I understand the later files have higher priority than the first one.

huard:
Can this be merged ?

aulemahal:
The logging workaround works fine, but I also made changes to the way errors are handled in the process handler. These change seem to be in opposition with concerns raised in #129, namely: Passing the error message as-is could potentially expose local paths and other stuff we usually do not want to share "publicly". 
I can remove that change and only merge the logging issue, or we can discuss it.

huard:
I understand the security concerns, but the status quo is not working for me. The WPS response often provides no information at all at what went wrong. Any code or directory structure that would be exposed will by definition be inside a conda environment that is publicly reproducible, so I really don't see what sensitive information can be exposed.

PR #174, opened 2021-04-09
Notebook showing subsetting and averaging.

### LINKED ISSUE #165: Add notebook demoing polygon averaging
## Description

Create small notebook using birdy to subset test netCDF dataset hosted on THREDDS.

Question: should we demo using a local hand-made polygon ? local polygon on file ? Polygon from WPS feature ?

aulemahal:
## Overview

This PR fixes #165.

Désolé David, I had this notebook ready in a local branch, but didn't push it since I wanted to do a bit more. I wanted to add handling of polygons without downloading them, but OWSlib is broken and I couldn't do it, so here it is without.

Changes:

* Added a notebook demonstrating subsetting with a bbox, with a polygon downloaded from WFS, demonstrating averaging with a small polygon and with a large polygon (with simplification made locally).


huard:
1. Do you have any idea why it doesn't work with ERA5 in my notebook ? 
2. Could you describe the issue you're having with the WFS command ?

aulemahal:
1. ERA5 on pavics is a HUGE dataset with humongus chunks,  it may be a simple uncaught timeout or something like that.
2. The WFS object defines a `getfeature` that retrieves (downloads) the feature. The object also defines a `getGETgetFeatureRequest` that returns the URL needed for that download. However, in owslib the WFS2.0 object is broken, filters cannot be used (geopython/OWSlib#710). And in the 1.0 and 1.1 objects, the `getGETgetFeatureRequest` is the same as on the 2.0 object (the broken one), but the `getfeature` method does work.

huard:
I get the same error with your notebook: 
```
ValueError: dimension lat on 0th function argument to apply_ufunc with dask='parallelized' consists of multiple chunks, but is also a core dimension. To fix, either rechunk into a single dask array chunk along this dimension, i.e., ``.chunk(lat: -1)``, or pass ``allow_rechunk=True`` in ``dask_gufunc_kwargs`` but beware that this may significantly increase memory usage.
2021-04-12 10:54:42,303 - werkzeug - INFO - 127.0.0.1 - - [12/Apr/2021 10:54:42] "POST / HTTP/1.1" 200 -
```
I suspect an unmet version requirement for clisops or xarray ?

aulemahal:
Is that not the chunking error that I fixed in #172?

aulemahal:
Oh! I just clicked! It's because we are using the pavics instance of finch, where the logging and rechunking fixes are not yet present. The notebooks and tests must be done against the local finch.

huard:
That's an error copied from my local server log.

huard:
To avoid HTML representation test failures, use # NBVAL_IGNORE_OUTPUT in the cell instead of returning the text format.

huard:
Argh... `make start` used the finch installed by pip, not the master... works now. Sorry about that. 

I think this is ready to merge, but I suggest we try to speed up the nb execution before doing so.

huard (approved):

aulemahal:
I removed the first plot (and data download), + added time subsetting to try to reduce the data size. The slowest cells are the average_polygons call and xESMF is in itself slower than the rest...

PR #175, opened 2021-04-12
Update CI for GitHub Workflows

Zeitsperre:


huard:
This seems to think it's a ruby package.

huard (approved):
There should be a `make test` as well, no ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Conda version
+        run: conda --version
+      - name: Create finch environment
+        run: conda env create -n finch-wps -f environment.yml
+      - name: Install and test finch-wps
+        run: |
+          source activate finch-wps
+          make develop
+          make start
+          sleep 2
```
Zeitsperre, line 42:
```suggestion
          sleep 2
          make test
```

PR #176, opened 2021-04-12
Release 0.7.3

huard:
## Overview

- update CHANGES with 0.7.3 release notes.
- Pin xesmf >= 0.5.3.


requirements.txt
```diff
@@ -2,6 +2,7 @@ cftime
 pywps==4.4
 xclim==0.24.0
 xarray
+xesmf>=0.5.3
```
tlvu, line 5: Houston, on Pypi https://pypi.org/project/xesmf/#history xesmf is only 0.3.0 !

huard: Yup. We're using https://github.com/pangeo-data/xESMF

huard: The core developer vanished, so we forked the repo and maintain it elsewhere.

tlvu: My point is that line will never work if xesmf on Pypi is stuck at 0.3.0.

huard: To be clear, xESMF cannot be installed with pip. It has to be installed with conda. 
However, once it's installed, my understanding is that we can still ask setup.py to check that the installed version meets some requirements, irrespective of how it was installed.

tlvu:
Testing building Finch docker from this branch with xESMF 0.5.3 conda package.

environment-docs.yml
```diff
@@ -10,3 +10,4 @@ dependencies:
 - ipython
 - xclim==0.24
 - matplotlib
+- xesmf>=0.5.3
```
tlvu, line 13: environment-docs.yml is only to build docs on ReadTheDocs.  Do we need xesmf to build docs (it wasn't there before)?  In general we want to keep this env file to the minimum to not burst the memory limit on ReadTheDocs.

aulemahal: I see you're right! I didn't realize we didn't install finch on RTD.

aulemahal, line 13:
```suggestion
```

huard:
@aulemahal ok to merge ? notebooks pass on my end.

aulemahal (approved):

huard:
Argh ! J'ai oublié bumpversion ! Je le fais sur master direct...

PR #189, opened 2021-05-28
Update xclim 0.27

aulemahal:
## Overview

This PR bumps xclim to 0.27 which was released soon.

Changes:

* Read in xclim variables from the new `xclim.core.utils.VARIABLES` dict. This kinda fixed a bug where variables not listed there would be added to ensemble processes as input, even if that doesn't make sense.


huard:
If you give me a couple of minutes, I'll try to complete the humidex PR.

huard:
J'étais mêlé, je pensais que c'était un PR xclim...

Zeitsperre:
RtD recently bumped to a major version. It could be that PyWPS requires `sphinx~=3.0` until that's addressed.

Zeitsperre:
Actually, this is getting weird; the function being imported was deprecated a long time ago. I get the sense that the PyWPS version is quite old. Is that possible?

Context: https://github.com/sphinx-doc/sphinx/blob/830b3fbe2babcc8df33f767ce3a406b16c0cac1c/CHANGES#L2060

aulemahal:
Oh! Haha, saw this : https://github.com/geopython/pywps/commit/56e9a8928505fa4406a50cfd62b91ca1a83d6dae

So, we are waiting for a pyWPS release then?

aulemahal:
Anyway, I'll need to update the env and reqs + remove the pip install from git when 0.27 comes out on pypi and conda.
Also, there's no rush, I might take a look at a few other issues next week.

huard:
PyWPS 4.4.4 has been released in early June. I'd like to merge this so we can work on the xclim 0.28 update.

huard:
Nb failures that I won't have time to fix. If someone can address that, would be great.

Zeitsperre:
PyWPS 4.4 seems to be complaining about a malformed netCDF, but the output looks fine. The stderr produced is what follows:

> syntax error, unexpected WORD_WORD, expecting SCAN_ATTR or SCAN_DATASET or SCAN_ERROR
 > context: <?xml^ version="1.0" encoding="UTF-8"?><ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd" version="1.0.0"> <ows:Exception exceptionCode="MissingParameterValue" locator="service" > <ows:ExceptionText>service</ows:ExceptionText> </ows:Exception></ows:ExceptionReport>

I can set nbval to ignore this, but it doesn't fix the failing docs. It seems to take issue with the new method of populating the ICCLIM indices. Not sure how best to tackle that.

Zeitsperre:
I should note that this error is NOT being raised in the CI tests, so the failure is because of the cell not producing any errors online. Go figure.

huard:
@Zeitsperre Please create new tickets if you feel that something's not behaving as it should, so we don't lose track over summer.

docs/source/notebooks/finch-usage.ipynb
```diff
@@ -64,7 +64,7 @@
       "    Parameters\n",
       "    ----------\n",
       "    tasmin : ComplexData:mimetype:`application/x-netcdf`, :mimetype:`application/x-ogc-dods`\n",
-      "        NetCDF Files or archive (tar/zip) containing netCDF files.\n",
+      "        NetCDF Files or archive (tar/zip) containing netCDF files. Minimum surface temperature.\n",
```
tlvu, line 67: This "Minimum surface temperature." text addition breaks Jenkins :(  New Finch 0.7.5 is good to be deployed to PAVICS?

PR #188, opened 2021-05-27
finch-usage.ipynb: update output due to new birdy 0.8.0

tlvu:
New birdy removed `output_formats`.

Clarification from DavidH: To clarify, the `output_formats` option still exists, but it's only appearing when the process has a complexoutput supporting more than one data format. 
https://github.com/bird-house/birdy/commit/2bea46c1aa20a91dfe38741476e202919c4b1ffa#diff-ba0f245ac7a3747697d42961f7a19568e294dc5d2b4ad14b62a47d8c240ed12eR230-R238

Also removed `NBVAL_IGNORE_OUTPUT` since Jenkins is configured to by
default ignore html changes, see PR https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/73.

Jenkins passing build with upcoming Jupyter env: http://jenkins.ouranos.ca/job/PAVICS-e2e-workflow-tests/job/new-docker-build-for-ravenpy/10/console

New Jupyter env PR: https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/75

To fix this error:
```
  _________ finch-master/docs/source/notebooks/finch-usage.ipynb::Cell 1 _________
  Notebook cell execution failed
  Cell 1: Cell outputs differ

  Input:
  help(wps.frost_days)

  Traceback:
   mismatch 'stdout'

   assert reference_output == test_output failed:

    'Help on meth...ut files.\n\n' == 'Help on meth...ut files.\n\n'
    Skipping 201 identical leading characters in diff, use -v to show
    - iable=None) method of birdy.client.base.WPSClient instance
    + iable=None, output_formats=None) method of birdy.client.base.WPSClient instance
    ?           +++++++++++++++++++++
          Number of days where daily minimum temperatures are below 0.

          Parameters
          ----------
          tasmin : ComplexData:mimetype:`application/x-netcdf`, :mimetype:`application/x-ogc-dods`
              NetCDF Files or archive (tar/zip) containing netCDF files.
          thresh : string
              Freezing temperature.
          freq : {'YS', 'MS', 'QS-DEC', 'AS-JUL'}string
              Resampling frequency.
          check_missing : {'any', 'wmo', 'pct', 'at_least_n', 'skip', 'from_context'}string
              Method used to determine which aggregations should be considered missing.
          missing_options : ComplexData:mimetype:`application/json`
              JSON representation of dictionary of missing method parameters.
          cf_compliance : {'log', 'warn', 'raise'}string
[...]

PR #191, opened 2021-06-16
test-notebooks target: nbval default ignore html and javascript output

tlvu:
With PR https://github.com/bird-house/finch/pull/188 removing `NBVAL_IGNORE_OUTPUT` since Jenkins is configured by
default to ignore html changes (PR https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/73),
test-notebooks target here is broken because it does not have the same
`conftest.py` as Jenkins.

This change brings the same `conftest.py` from Jenkins here to fix
test-notebooks target.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,3 @@
+def pytest_collectstart(collector):
+    if collector.fspath and collector.fspath.ext == '.ipynb':
+        collector.skip_compare += 'text/html', 'application/javascript',


tlvu:
@Zeitsperre do you know why Github actions is stuck at xarray 0.17.0?  Still have this error `pywps.app.exceptions.ProcessError: NoApplicableCode: Process error: Finch failed with AttributeError dict object has no attribute collection_annotations`.

According to @aulemahal xarray 0.18.2 should be good and that version is available now.

huard:
Ok, go ahead and merge, we'll try to fix the errors in the next PR.

PR #202, opened 2021-09-22
Multiple rcps for ensemble processes

### LINKED ISSUE #192: Add support for multi-RCP inputs
Add support for multi-RCP inputs (eg: `rcp85, rcp45`).

aulemahal:
## Overview

This PR fixes #192 

Changes:

* I changed the `rcp` input to accept up to 3 "occurs"
* In the common ensemble handler, it simply iterates over the rcps and does the whole process serially. 
* If a single rcp is requested, the output is unchanged. If there are multiple, a new "rcp" dimension appears.

Missing tests.


aulemahal:
Tests added.

When the input files do not have the rcp in their name, that creates name conflicts in the work directory. To work around this, I create a sub directory for each rcp and set the process' workdir in the loop. As it is now, the problem only occurs in the tests...

Note that the `rcp` option must be passed _multiple_ times, and not as a comma-separated string (as for percentiles). In `birdy` this means passing a list.

finch/processes/ensemble_utils.py
```diff
@@ -527,73 +530,87 @@ def ensemble_common_handler(process: Process, request, response, subset_function
...
 
-    write_log(process, "Computing indices", process_step="compute_indices")
+        indices_files = []
 
-    input_groups = make_indicator_inputs(
-        process.xci, request_inputs_not_datasets, subsetted_intermediate_files
-    )
-    n_groups = len(input_groups)
+        warnings.filterwarnings("ignore", category=FutureWarning)
+        warnings.filterwarnings("ignore", category=UserWarning)
```
huard, line 580: Could you use a context instead ?
E.g. `with warnings.catch_warnings():`

huard (approved):

finch/processes/ensemble_utils.py
aulemahal: haha yes I could, but that's not my code. Simply github not understanding all the diff.

matprov (approved):

finch/processes/ensemble_utils.py
aulemahal: After reading on `catch_warnings` it seems it doesn't do much more than what we have. It would be replacing 4 lines by 3+indentation, doesn't feel like a victory. Also it is said to be unsafe in multithreading settings...

PR #211, opened 2021-11-10
run load() before switching to single-threaded scheduler

huard:
## Overview

This PR fixes #204

### DIFF
--- a/finch/processes/utils.py
+++ b/finch/processes/utils.py
@@ -657,6 +657,9 @@ def dataset_to_netcdf(
         for v in ds.data_vars:
             encoding[v] = {"zlib": True, "complevel": compression_level}
 
-    # This is necessary when running with gunicorn
+    # Perform computations
+    ds.load()
+
+    # This is necessary when running with gunicorn to avoid lock-ups
     with dask.config.set(scheduler="single-threaded"):
         ds.to_netcdf(str(output_path), format="NETCDF4", encoding=encoding)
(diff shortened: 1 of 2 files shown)


finch/processes/utils.py
```diff
@@ -657,6 +657,9 @@ def dataset_to_netcdf(
         for v in ds.data_vars:
             encoding[v] = {"zlib": True, "complevel": compression_level}
 
-    # This is necessary when running with gunicorn
+    # Perform computations
+    ds.load()
+
+    # This is necessary when running with gunicorn to avoid lock-ups
     with dask.config.set(scheduler="single-threaded"):
         ds.to_netcdf(str(output_path), format="NETCDF4", encoding=encoding)
```
aulemahal, line 665:
```suggestion
    ds.to_netcdf(str(output_path), format="NETCDF4", encoding=encoding)
```
If the data is loaded, dask is not involved anymore, and the config set has no effect.

aulemahal (approved):
I believe indeed this solution will be better for performances!

finch/processes/utils.py
```diff
@@ -657,6 +657,8 @@ def dataset_to_netcdf(
         for v in ds.data_vars:
             encoding[v] = {"zlib": True, "complevel": compression_level}
 
-    # This is necessary when running with gunicorn
-    with dask.config.set(scheduler="single-threaded"):
-        ds.to_netcdf(str(output_path), format="NETCDF4", encoding=encoding)
+    # Perform computations
+    ds.load()
+
+    # This is necessary when running with gunicorn to avoid lock-ups
```
cjauvin, line 663: this comment should be removed

PR #212, opened 2021-11-16
Prep 0.7.6 release

huard:


aulemahal (approved):
Don't we need to change `setup.cfg` and `setup.py` too?

huard:
And that's why reviews are useful !

PR #221, opened 2022-01-20
Process to convert Geoseries to netCDF + Doc improvements

### LINKED ISSUE #208: Process for converting GIS vector data to netCDF
## Description

A process that takes in a GIS url (ex. geojson), reads it with `geopandas`, converts the geometries to CF-compliant variables with `cf-xarray` and cleans it up to return a dataset along `time`, as normally expected by the other processes.

aulemahal:
## Overview

This PR fixes #208

Changes:

* Added `GeoseriesToNetcdfProcess` to convert Geospatial (time)series into CF-compliant netCDFs.
* Changed readthedocs build mechanism from `conda` to `pip` for a fast non-crashing behaviour.
    - Modified the mock imports in `conf.py` to fix the parsing and documentation of processes.

It takes a geojson URL and passes through `cf_xarray.geometry.shapely_to_cf` to convert to a CF dataset.

- We try to convert variables to datetime, int or float. Otherwise they are left as strings.
- We can control which column becomes the series index. With geomet, it's "LOCAL_DATE", which is the default.
- We can give a variable which becomes the index over features (points). If not given, the dim is "features" and has a simple integer coordinate.
- We can try to squeeze all variables along either of the axes. For example, with geomet, this makes "CLIMATE_IDENTIFIER"  a variable only along "features", as the identifier doesn't change in time.

## Additional Information
Example:
```python3
out = wps.geoseries_to_netcdf(
  "https://api.weather.gc.ca/collections/climate-daily/items?datetime=2010-03-01%2000:00:00/2010-06-02%2000:00:00&bbox=-115,50,-112,51&sortby=PROVINCE_CODE,STN_ID,LOCAL_DATE&f=json&limit=150000&startindex=0",
  index_dim="LOCAL_DATE",
  feat_dim="STATION_NAME",
  squeeze=True,
)
```

Gives:
```
<xarray.Dataset>
Dimensions:                   (STATION_NAME: 23, LOCAL_DATE: 94)
Coordinates:
  * STATION_NAME              (STATION_NAME) object 'TRAVERS AGCM' ... 'BARON...
  * LOCAL_DATE                (LOCAL_DATE) datetime64[ns] 2010-03-01 ... 2010...
    lon                       (STATION_NAME) float64 ...
    lat                       (STATION_NAME) float64 ...
Data variables: (12/34)
    id                        (STATION_NAME, LOCAL_DATE) object ...
    CLIMATE_IDENTIFIER        (STATION_NAME) object ...
    ID                        (STATION_NAME, LOCAL_DATE) object ...
    PROVINCE_CODE             (STATION_NAME) object ...
    LOCAL_YEAR                float64 ...
    LOCAL_MONTH               (LOCAL_DATE) float64 ...
    ...                        ...
    HEATING_DEGREE_DAYS_FLAG  (STATION_NAME, LOCAL_DATE) object ...
    MIN_REL_HUMIDITY          (STATION_NAME, LOCAL_DATE) float64 ...
    MIN_REL_HUMIDITY_FLAG     (STATION_NAME, LOCAL_DATE) float64 ...
    MAX_REL_HUMIDITY          (STATION_NAME, LOCAL_DATE) float64 ...
    MAX_REL_HUMIDITY_FLAG     (STATION_NAME, LOCAL_DATE) float64 ...
    geometry_container        float64 ...
```


finch/processes/wps_geoseries_to_netcdf.py
```diff
@@ -0,0 +1,200 @@
...
+                "Geospatial series",
+                abstract="Series of geographical features.",
+                supported_formats=[FORMATS.GEOJSON],
+                min_occurs=1,
+                max_occurs=1,
+            ),
+            LiteralInput(
+                "index_dim",
+                "Index dimension",
+                abstract="Name of the column in the data to be converted into the index dimension.",
```
huard, line 32: This should usually be time, right ? Eventually we could try to autodetect that column based on which can be converted to a datetime object.

finch/processes/wps_geoseries_to_netcdf.py
```diff
@@ -0,0 +1,200 @@
...
+                data_type="string",
+                min_occurs=1,
+                max_occurs=1,
+            ),
+            LiteralInput(
+                "feat_dim",
+                "Feature dimension",
+                abstract="Name of the column in the data to be converted into the 'feature' dimension. Each geometry must be different along and only along this dimension.",
+                data_type="string",
+                default="",
```
huard, line 42: Isn't that always "geometry" ?

finch/processes/wps_geoseries_to_netcdf.py
```diff
@@ -0,0 +1,200 @@
...
+            LiteralInput(
+                "squeeze",
+                "Squeeze variables",
+                abstract="Squeeze variables that are replicated along one of the dimensions.",
+                data_type="boolean",
+                default=False,
+                min_occurs=0,
+                max_occurs=1,
+            ),
+            LiteralInput(
```
huard, line 55: I suggest commenting this, and add it later when more than one option is available, setting the default to the current value. This way, no code will break.

finch/processes/wps_geoseries_to_netcdf.py
```diff
@@ -0,0 +1,200 @@
...
+                ],
+            ),
+            wpsio.output_log,
+        ]
+
+        super().__init__(
+            self._handler,
+            identifier="geoseries_to_netcdf",
+            version="0.1",
+            title="Convert a geospatial series to a CF netCDF.",
```
huard, line 83:
```suggestion
            title="Convert a geospatial series to a CF-compliant netCDF.",
```

finch/processes/wps_geoseries_to_netcdf.py
```diff
@@ -0,0 +1,200 @@
...
+                            raise
+
+        write_log(self, "Geoseries converted to CF dataset", process_step="converted")
+
+        host = urlparse(geo_url).netloc
+        if host is None:
+            source = "geospatial series data"
+        else:
+            source = f"data downloaded from {host}"
+        ds.attrs["history"] = update_history(
```
huard, line 155: Add timestamp.

finch/processes/wps_geoseries_to_netcdf.py
Zeitsperre: Usually. If the GeoSeries was built from a SQL request, the convention is to name that column "the_geom".

finch/processes/wps_geoseries_to_netcdf.py
aulemahal: In most cases yes. I guess that timeseries will only have a single column convertible into datetime. I also guess that in the case of the column convertible to datetime it is _always_ the index of the timeseries (the guess has to work in both directions).

finch/processes/wps_geoseries_to_netcdf.py
aulemahal: That's not what I meant here, so I need to clarify the docs. 
When opening a geojson with geopandas and converting it to xarray, the variable containing the shapes is _always_ named "geometry" (regardless of what came in geopandas). After the reshape, the dimension is named "features".  And that dimension has a basic integer coordinate.

What the users are specifying here is the name of another column in the input table that could be used as the coordinate of the "features" dimension (and it also renames the dimension).

huard: Got it.

finch/processes/wps_geoseries_to_netcdf.py
aulemahal: `update_history` already does that! I also specifies the xclim version, but we are not using xclim here... Is this a problem? Should we reimplement a version for finch?

tlogan2000:
> MDR, RTD failed with a memory error... Can someone retrigger it to try again?

Not sure what you mean here

aulemahal:
> Not sure what you mean here
Readthedocs seems to have difficulties with finch. The environment solving seems heavy.

finch/processes/wps_geoseries_to_netcdf.py
tlogan2000: Is it possible that the geojson would have three separate columns (e.g. year, month, day) that the service should concatenate into one?    In this case the input could be a list of dims?  Not sure if possible but just thinking bigger picture than just what geomet give us

tlogan2000:
> @tlogan2000 What do you think? (and who at CRIM should I ping?)

I think it seems pretty good made a comment above about the index column.  Was there something in the notebook to review? 
Does this reproduce the results as in @huard 's  new notebook for pavics-sdi?

finch/processes/wps_geoseries_to_netcdf.py
aulemahal: Oh boy, I guess you're right! We could have:
If a string : do as it does now
If a list of 3 strings : assume it means year-month-day
if not given : use first column that converts to datetime.

Seems we are trying to cover very specific cases here, how robust would it be?

tlogan2000:
People at CRIM who will need to know about this: @matprov and maybe Louis-David Perron (doesn't seem to be in the finch collabs at the moment)

aulemahal:
> Does this reproduce the results as in @huard 's new notebook for pavics-sdi?
It adds the possibility to squeeze variables when possible, to cast columns whenever possible and to assign a coordinate to the "feature" dimension.
But it doesn't put any attributes on the output (it has no way to know), doesn't mask variables according to their "flag" (also quite specific to geomet) and doesn't reindex the final series on a continuous timeseries (because it doesn't actually assume it is a timeseries).

finch/processes/wps_geoseries_to_netcdf.py
tlogan2000: I think I would say "let's cross that bridge if and when we get to it"  ... was just thinking out loud really

finch/processes/wps_geoseries_to_netcdf.py
huard: oh, right... 

If this is going to go in prod for ECCC, then yes.

environment-docs.yml
```diff
@@ -12,3 +12,10 @@ dependencies:
 - matplotlib
 - birdy > 0.8.0 # >= pin fails because conda package seems to be tagged v0.8.1 instead of just 0.8.1.
```
Zeitsperre, line 13:
```suggestion
- birdy > 0.8.1
```
This issue has been fixed.

aulemahal:
@huard @Zeitsperre This PR changes the readthedocs method from conda to pip. Conda as the problem of being very slow when we use `conda-forge` (which we need to). `mamba` did not completely solve the problem, it was still very slow (but didn't crash). using `pip` is quite fast (140s total) and I don't see any downsides for now?

aulemahal:
@huard @Zeitsperre @tlvu @tlogan2000 and I don't know who else needs to know:
FYI: It seems that sphinx 4.4 (or even earlier on) is now throwing _errors_ instead of warnings when a title is of an inconsistent level. In notebooks this happens when a "### title" follows a "# title", without any "## title" in between.

Also, rtd uses sphinx 1.8 with pip is the project is older than october 2020. A bit confusing, but we need to pin it ourselves to override this. The last commit upgrades sphinx, fixes a title inconsistency error and updates the process list (see upcoming issue).

aulemahal:
I don't thinnk it's a notion of a RTD image, more that the 2.0 sphinx update must have been highly breaking and RTD didn't want to break all their user's workflows. Pinning our own version of sphinx is the recommended way in the official doc!

tlvu:
> but there's still a bug in the doc build: the processes can't be parsed

Is this related to the various mock imports?  Is switching from conda to pip install okay with the mock imports?

tlvu:
> > but there's still a bug in the doc build: the processes can't be parsed
> 
> Is this related to the various mock imports? Is switching from conda to pip install okay with the mock imports?

Processes seem to be generated fine here: https://pavics-sdi--221.org.readthedocs.build/projects/finch/en/221/processes.html

Screenshot from 2022-01-21 21-05-05

aulemahal:
As pointed out in slack, the last commit fixed the doc problem. We have thus switched to a pip install on RTD and I reorganized the mock imports to fix the issue. As we _need_ xclim in the doc env, we install many of the dependencies, so I simply removed a lot of them from the mock imports. 
I just realized the way to specify the python version has changed in the RTD config, I'll push that as a last change.

PR #225, opened 2022-02-03
Adapt to pandas 1.4

aulemahal:
## Overview

This PR fixes an issue in the DAP subset notebook that seems to be caused by the new pandas (1.4)
The fix is retro-compatible.


docs/source/notebooks/subset.ipynb
```diff
@@ -18,10 +18,10 @@
    "id": "8f4fd0c3",
```
tlvu, line 18: <p>Is it possible to silence this kind of warnings "/opt/miniconda3/envs/finch/lib/python3.9/site-packages/geopandas/io/file.py:362: FutureWarning: pandas.Int64Index is deprecated and will be removed from pandas in a future version. Use pandas.Index with the appropriate dtype instead."</p><p> </p><p>This will break Jenkins.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/bird-house/finch/pull/225/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

aulemahal: Not directly, this has to be done in `geopandas`. but we could filter out any "FutureWarning" I guess.

aulemahal:
@tlvu  I added a cell at the top of the notebook that filters out `FutureWarning` and that is hidden from the final doc (nbsphinx hides it).
I also tried to move the "NBVAL Ignore Output" comments to the cell's metadata, as documented here: https://nbval.readthedocs.io/en/latest/#Using-tags-instead-of-comments.  If that works, it would be cleaner for the reader?

docs/source/notebooks/dap_subset.ipynb
```diff
@@ -508,17 +508,17 @@
...
+     "iopub.status.idle": "2022-02-04T18:10:43.902766Z",
+     "shell.execute_reply": "2022-02-04T18:10:43.900886Z"
     }
    },
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "{'lat': 1, 'lon': 2, 'time': slice(5040, 6840, None)}"
+       "{'lat': array(1), 'lon': array(2), 'time': slice(5040, 6840, None)}"
```
tlvu: Should add a NBVAL_IGNORE_OUTPUT, else this cell will break in Jenkins that use a Jupyter env that do not have panda 1.4.0 yet.

docs/source/notebooks/subset.ipynb
```diff
@@ -109,15 +128,18 @@
...
-     "iopub.status.idle": "2022-01-24T17:53:50.956633Z",
-     "shell.execute_reply": "2022-01-24T17:53:50.955914Z"
-    }
+     "iopub.execute_input": "2022-02-04T18:11:35.479101Z",
+     "iopub.status.busy": "2022-02-04T18:11:35.477269Z",
+     "iopub.status.idle": "2022-02-04T18:11:38.681044Z",
+     "shell.execute_reply": "2022-02-04T18:11:38.682228Z"
+    },
+    "tags": [
+     "nbval-ignore-output"
```
tlvu, line 141: whooo, fancy !, I'll test this in a Jenkins run to be sure.

When we simply open the notebook, do we see this tag?  Or it is hidden so unless we search, we do not know which cells have this tag and which do not.  Could potentially be a hidden debugging problem later.  But yeah we can try.

tlvu (approved):
LGTM, just make sure the  changes are backward compat with currrent Jupyter env that  do not have pandas 1.4

I can help test on Jenkins.

tlvu:
Using notebook "tags" works http://jenkins.ouranos.ca/job/PAVICS-e2e-workflow-tests/job/master/1463/console

Only error left is to also ignore the output change for this cell to be backward compat.
```
14:47:58  ______ finch-adapt-pd-1.4/docs/source/notebooks/dap_subset.ipynb::Cell 6 _______
14:47:58  Notebook cell execution failed
14:47:58  Cell 6: Cell outputs differ
14:47:58  
14:47:58  Input:
14:47:58  index
14:47:58  
14:47:58  Traceback:
14:47:58   mismatch 'text/plain'
14:47:58  
14:47:58   assert reference_output == test_output failed:
14:47:58  
14:47:58    "{'lat': arra... 6840, None)}" == "{'lat': 1, '... 6840, None)}"
14:47:58    - {'lat': 1, 'lon': 2, 'time': slice(5040, 6840, None)}
14:47:58    + {'lat': array(1), 'lon': array(2), 'time': slice(5040, 6840, None)}
14:47:58    ?         ++++++ +         ++++++ +
```

docs/source/notebooks/subset.ipynb
aulemahal: On the jupyter notebook interface: View -> Cell toolbar -> Edit Metadata. And then on each cell a "edit" button appear. When clicked it shows the JSON part with that tag.
I agree it isn't evident and might make debugging harder.

docs/source/notebooks/dap_subset.ipynb
```diff
@@ -1881,17 +1882,17 @@
...
+     "iopub.status.idle": "2022-02-04T18:10:46.164347Z",
+     "shell.execute_reply": "2022-02-04T18:10:46.166583Z"
     }
    },
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "{'lat': 1, 'lon': 2, 'time': slice(5040, 6840, None)}"
+       "{'lat': array(1), 'lon': array(2), 'time': slice(5040, 6840, None)}"
```
tlvu: @aulemahal Missed ignore output on this one as well !  I'll do it so I can try this tags thing.

PR #226, opened 2022-02-07
Release 0.8.2

aulemahal:
Prepare for release 0.8.2, I hope it works this time...


tlvu:
@aulemahal Since `dask` and `distributed` are not pinned by `xclim`, would you need to pin them here for Finch performance to not degrade?

aulemahal:
Indeed, I don't think xclim is a good place to pin them as it is a _library_, but finch sure is since it is an _application_.

tlvu:
> Indeed, I don't think xclim is a good place to pin them as it is a _library_, but finch sure is since it is an _application_.

Yes in principle xclim is not a good place to pin since this is a performance problem only, not incompatibility problem.

However, in practice, it will save a lot of trouble for a lot of downstream users.  Everyone that update to the new xclim will have to know to pin dask and distributed if they do not want to suffer a performance problem, a lot of duplicate effort.  They might even falsely accuse xclim of the problem, like I originally did, while in fact it is dask and distributed.

So hope my reasoning is clear.  But it's your call.

huard (approved):

tlvu:
@aulemahal 
I just realized the pinning of `dask` and `distributed` should be done in the `environment.yml` instead because `Dockerfile` do not use the `requirements.txt` file !

Is there a similar code as the homepage notebook 3 we can try to see if Finch performance is affected similarly?  Homepage notebook 3 was hitting xclim directly.  We need to go through Finch.

PR #233, opened 2022-04-19
release 0.8.3: proper pin of dask and distributed for the Finch docker image

tlvu:
Pin `dask` and `distributed` for the performance problem observed in https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/issues/100.

This is the same pin as commit https://github.com/bird-house/finch/pull/226/commits/7bf4a7e1bcb12bc88d15bdcbbd484e81106be67c but in file `environment.yml` for Docker build.

Our nightly Jenkins has also been consistently falling at the notebook `subset-user-input.ipynb` with the following error and leave scrap in the WPS queue:
```
21:01:21  ServiceException: <?xml version="1.0" encoding="utf-8"?>
21:01:21  <ExceptionReport version="1.0.0"
21:01:21      xmlns="http://www.opengis.net/ows/1.1"
21:01:21      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
21:01:21      xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd">
21:01:21      <Exception exceptionCode="NoApplicableCode" locator="NotAcceptable">
21:01:21          <ExceptionText>Request failed: (&#x27;Connection aborted.&#x27;, RemoteDisconnected(&#x27;Remote end closed connection without response&#x27;))</ExceptionText>
21:01:21      </Exception>
21:01:21  </ExceptionReport>
```

Will perform release procedure at https://github.com/bird-house/finch/blob/master/docs/source/dev_guide.rst#bump-a-new-version once this PR approved.


huard (approved):

tlvu:
Having problems building on DockerHub with error ` picosat: out of memory in 'resize'` so while investigating, I notice xclim is pinned at `0.31`.  Given current xclim is at `0.35`, is it normal that xclim is pinned so old?  The previous Finch 0.8.2 is also using this old xclim, while probably xclim `0.34` exists, is that expected?

https://github.com/bird-house/finch/blob/56ab1639a4e85aec3be7d635009382fa02fe2cef/environment.yml#L24

aulemahal:
Updating xclim within finch is usually more complex than only bumping the number, so we pin a specific version each time. And in the last weeks none of us had the time and the need to update finch.

PR #235, opened 2022-06-20
Consistent output_format and output_name

### LINKED ISSUE #231: Feature request: Add output_format parameter for processes
## Description

For the Analysis on AHCCD stations data, we need to be able to provide csv and netcdf output format for xclim indicators processes. Currently, only netcdf is provided.

### LINKED ISSUE #234: "output_name" broken for indicators
## Description

"output_name" gets listed as an input by the indicator WPS processes, but when given those processes fail with "TypeError got an unexpected keyword argument output_name".  The error goes back to xclim : the problem is that the argument is passed to `compute_indices` and then to the indicator. But it's a finch-only think, xclim doesn't recognize it and fails.

## Environment

* Finch version used, if any: 0.8.2, PAVICS andmaster
* Python version, if any: 3.9 and maybe others
* Operating System: Linux

## Steps to Reproduce

```
import os
import xarray as xr
from birdy import WPSClient

pavics_url = "https://pavics.ouranos.ca/twitcher/ows/proxy/finch/wps"
url = os.environ.get("WPS_URL", pavics_url)
verify_ssl = True if "DISABLE_VERIFY_SSL" not in os.environ else False
wps = WPSClient(url, verify=verify_ssl)

tasmin = "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/testdata/flyingpigeon/cmip3/tasmin.sresa2.miub_echo_g.run1.atm.da.nc"

o = wps.tn_days_below(
    tasmin=tasmin
    thresh="-5 degC",
    freq="YS",
    check_missing="pct",
    output_name="an_output_name",
)
```

aulemahal:
## Overview

This PR fixes #231 and fixes #234.

Third point is a breaking change!

## Changes:

*  `output_name` should work wherever it is added. 
* xclim indicators now accept `output_format` which can be "netcdf" or "csv". In the latter case, a zip file is returned,  with the data in a csv and the metadata in a txt.
* The main output was renamed `output`, instead of `output_netcdf`. This change was propagated to all other processes, so the main output should now always be `output`.


aulemahal:
@perronld, what do you think? Beware the breaking change in the output name.

huard (approved):

perronld:
> @perronld, what do you think? Beware the breaking change in the output name.

I can't forsee anything that would break on the portal itself, since it doesn't care of the output variables, only the input. 
If a user's script gets broken because of this change, it will be a very easy fix.

aulemahal:
@Zeitsperre @tlvu You might see from the commit that I had to fight the CI today, However, you might be interested in this neat solution for mamba-based github actions workflows! The step "provision-with-micromamba" automatically builds and activates an env from `environment.yml`. (but you have to execute the following code with `shell: bash -l {0}` for it to work.
Also, with `cache-downloads: true` github caches the downloads between runs (however within a branch). This accelerates the CI by a significant amount!

requirements.txt
```diff
@@ -14,7 +13,7 @@ bottleneck
 netcdf4
 numpy
 unidecode
-dask[complete]<=2022.1.0
+dask[complete]
```
tlvu, line 16: @aulemahal the performance problem with dask is fixed?

Issue #223, opened 2022-01-24
EQM and average_shape tests fail randomly


aulemahal:
## Description

The EQM and average_shape tests of the test suite sometimes fail randomly. 

The EQM test seems to fail because the random inputs yields an adjusted series slightly too different than the expectation. Raising the relative tolerance of the assertion could solve this easily.

The  `average_shape` test fails with ` ValueError: ESMC_GridCreateNoPeriDim() failed with rc = 545.    Please check the log files (named "*ESMF_LogFile").`, which seems related to a misformed grid. However, I have been unable to replicate this bug locally, I have no idea where it comes from...

## Environment

* Finch version used, if any: master
* Python version, if any: build , so 3.8
* Operating System: github actions (ubuntu?)

aulemahal:
EQM test relaxed in 947ce2a.

PR #236, opened 2022-06-21
Update xclim to 0.37

### LINKED ISSUE #219: Upgrade xclim to 0.32

aulemahal:
## Overview

This PR fixes #219

Changes:

* In xclim, all percentile threshold variables got their name changed to something more generic. Ex: `tn10` to `tasmin_per`. That means the information about the default percentile was lost in some cases. In this branch, those inputs are transformed into a `perc_tasmin` input, accepting an int, so that the percentile can be chosen by the user. The default value comes from an hardcoded list in `finch/processes/constants.py`. This only applies to ensemble processes. For basic indicators, only the input name changed.
* Modified the way the "accepted" variables were parsed from xclim.


huard (approved):

aulemahal:
CI failures are #223 and it is tapping me on nerves... I'll try to just change the whole notebook cell to avoid (?) the issue because I can't understand it.

PR #237, opened 2022-06-21
Avoid ESMF errors in notebook

### LINKED ISSUE #223: EQM and average_shape tests fail randomly

aulemahal:
## Overview

This PR fixes #223 
I made two changes in `subset.ipynb`, I'm not sure which one will fix the bug (if the bug is fixed).

* Used a real temporary directory for saving polygon files.
* Added `preserve_topology=True` and `buffer(0)` to the simplification of shapes.


Zeitsperre (approved):
LGTM

PR #238, opened 2022-06-22
Fix naming when subsetting OpenDAP data

### LINKED ISSUE #228: Processes using two variables (tasmax + tasmin) such as heat_wave_frequency fails
## Description

When running ensemble_grid_point_heat_wave_frequency, the process ends with an exception: KeyError 'tasmax' on finch/processes/ensemble_utils.py in compute_intermediate_variables at line 469.

After debbuging, I noticed that make_file_groups doesn't work as expected because all the files in files_list has a tempfile.mkstemp characters in its name, so the grouping is not being performed. The mkstemp is being called by resource._build_file_name(resource.url) because the output files are being created (with zero bytes) in the work directory before the subset is done. 

## Environment

* Finch version used, if any: birdhouse/finch:version-0.8.2 docker image

## Steps to Reproduce

```
Run ensemble_grid_point_heat_wave_frequency with finch deployed to use threeds as data source.
```

aulemahal:
## Overview

This PR fixes #228.

Changes:

* Remove use of `ComplexInput._build_file_name` to avoid randomness in filenames. Replaced with the same logic, adapted to finch.
* Replaced `resource.file` calls in logging with `getattr(resource, resource.prop)`, to avoid downloading files for no reason.


aulemahal:
The only risk is if two input files have the exact same name.

huard:
Could we raise an error if that happens, to avoid returning wrong outputs?

aulemahal:
Oh! I just realized there is not bug here. 

When using a normal xclim process, if two input files have the same name, this `ComplexInput._build_file_name` actually avoids the problem before finch even tries to open the input. 

When using an ensemble process, the user doesn't give their own file. So you may have files with identical names only if you  give them in the dataset config. I didn't test what happens in that case, but that's easily avoidable when configuring finch.

PR #250, opened 2022-07-20
Removed mandatory chunking for local files

perronld:
## Overview

This PR fixes performance issue in production for climatedata.ca 

Changes:

* Removed mandatory chunking when opening local netcdf files

## Additional Information

This change will most likely be challenged since I don't know its consequence on other projects.
With this change, subset_grid_point_dataset (single cell) takes less than 5 seconds instead of 15 minutes, with a full BCCAQv2 dataset compressed and chunked to (lat=1,lon=1,time=len(time)).  (Magnetic storage, warm cache). Memory usage also stays very low during the process.


aulemahal:
I'm not sure how this work?
Isn't the fact that the netCDF data is chunked irrelevant when we use `open_dataset`? I thought xarray wasn't able to replicate netCDF chunks?
This means that all the data has to fit in the memory for this to work. Do you have enough RAM for the whole BCCAQv2?

aulemahal (commented):
There some problems here. Most of them go back to the inability of xarray to guess chunk sizes from the netCDF chunks. There's a long lived issue on xarray, pydata/xarray#1440.

The problem here is that if this accelerates a single gridpoint subset, that's because xarray was able to perform the subset before loading anything, but for any other non-subsetting operation, this will just make finch implode.

The subsetting operations are slow in part because:
- The automatic chunks sizes are not adequate
- The chunking happens _after_ the opening

For the gridpoint operation, we could simply avoid using dask because we already know only one point will be used. However, what if someone request the whole map through `bbox`? A problem here, is we don't know the number of points before actually performing the subset, so this information is unavailable when opening the dataset.

I'm pushing a suggestion, but that removes any ameliorations for `bbox`... I'm not "requesting changes" because I'm leaving on vacation and don't want to block this.

Would it make sense to have a way of passing information about the chunking to use specific for the given file? Since there is a kinda frozen list of files, may be this could help?

Or use Zarr :sweat_smile:

aulemahal:
I note that `chunk_dataset` is currently kinda useless, I'm not sure what was my plan with it anymore... Also, the SDBA processes do not ensure the time dimension isn't chunked, which isn't great... ach.

perronld:
I'm currently running this branch in our staging environment (with the whole dataset as local files),  and subset performances are excellent for both grid (one or more cells), and bbox.

finch/processes/utils.py
```diff
@@ -286,37 +286,39 @@ def drs_filename(ds: xr.Dataset, variable: str = None):
 def try_opendap(
     input: ComplexInput,
     *,
-    chunks=None,
+    chunks='auto',
```
perronld, line 289: This changes unfortunately breaks analysis on AHCCD stations datasets
```NotImplementedError: Can not use auto rechunking with object dtype. We are unable to estimate the size in bytes of object data```

aulemahal: I added a try..except for that case!

aulemahal:
Build failures were fixed in #258. So we can wait for this branch to be merged.

aulemahal:
@matprov  Do you need this in 0.10 ?

matprov:
@aulemahal Yes please! I've tested this branch after your last commit and no regression on staging env.

PR #259, opened 2022-09-30
Add pre commit and blacken/cleanup code base

Zeitsperre:
## Overview

Changes (to date):

* Added a pre-commit configuration (largely based on that of `xclim`)
* Created yaml, json, and python code style configuration entries
* Ran `flake8` and `isort` hooks over the code base
* Ran NBstripout to strip execution metadata
* Reformatted many docstrings to adhere to pycodestyle
* Blackened code style
* Added pycodestyle `noqa` codes where needed
* Updated GitHub Actions versions

## Related Issue / Discussion

This is a draft first effort to adopt pre-commit hooks in Finch. This PR should not be a priority until the PR queue is empty, as the changes will be incompatible with existing proposed changes.

The PR should also serve as an opportunity to adopt numpy convention docstrings, which will need to be converted manually.

The end result should be a development workflow and coding conventions that are very similar to other existing Ouranos projects.

## Additional Information


.cruft.json
```diff
@@ -1,34 +1,34 @@
 {
-    "template": "https://github.com/bird-house/cookiecutter-birdhouse",
```
tlvu, line 2: I think cruft writes this file.  Maybe exclude it from the formatter as well.  Could leave this here as well, up to you.  We'll see at the next cruft update if cruft will revert this change or able to keep it.

tlvu (approved):
All good for me as well.  I am not sure about my comment for the cruft.json file.  Up to you.

tlvu:
Just curious, could we trigger all those formatter locally or they can only be triggered on github?

Zeitsperre:
> Just curious, could we trigger all those formatter locally or they can only be triggered on github?

`$ pre-commit run --all-files`

.cruft.json
Zeitsperre: I don't exclude it in other projects. So long as the order of entries doesn't get sorted, it should be fine to let the JSON formatter manage whitespace.

PR #258, opened 2022-09-26
Generalize ensemble datasets

aulemahal:
## Overview

This PR fixes #247, fixes #251.

Changes (breaking changes in bold):

* Rewrote all the management of datasets so that everything is parsed from a configuration file (`datasets.yml`).
* **Renamed the arguments to match the change: `rcp -> scenario`, `dataset_name -> dataset`.** (The latter was broken anyway).
* **Support for `candcs-u5` and `candcs-u6`**.  The name "bccaqv2" is translated into  `candcs-u5`.
* Because the CanDCS-U6 files have a deeper structure than the flat CanDCS-U5 list, the parsing methods have been generalized to recursively dig into subfolders (up to a given depth).
* The "allowed_values" of the WPS processes (args: `scenario, models`) are the union of all values of all datasets. However, the computation function will raise a "ValueError" if it receives arguments invalid for the chosen dataset. This means that currently, two kinds of errors are raised even though it's the same principle, I'll look into that.
* In theory, we could add any ensemble dataset even if they include more variables than `tasmax, tasmin, pr`. The code is currently generalized and the presence of new variables should trigger the wrapping of the additional compatible indicators,
* **Output filenames will be prefixed by the dataset name.** (unless a custom name is requested)
* The selection function accepts a list of string for `models`, but also a list of tuples (model name, realization). This is used for the PCIC12 "sublist", but we could make it visible to the user?
* There aren't global variable lists anymore.  The available variables are dataset-dependent and so are the compatible indicator. The ensemble common handler checks if the requested indicator is compatible with the request dataset. 
     - I removed `finch.processes.xclim` as it made no sense with this change.
* Removed `finch.processes.constants` because there was only 1 constant left.

## Related Issue / Discussion
@tlogan2000  Is there a sub-list for CanDCS-U6, similar to PCIC12?

## Additional Information
~I still need to update the doc, I'll add a section on how to pass dataset configuration.~

@perronld For local files, the YAML will look like:
```yaml
candcs-u5:
  local: true
  depth: 0
  path: /path/to/the/dataset/
  suffix: "*.nc"  # This is only used for local files
  pattern: "{variable}_{frequency}_BCCAQv2+ANUSPLIN300_{model}_{scenario}_{realization}_{date_start}-{date_end}.nc"
#  [the rest is the same]
```
The tests use a specific yaml that works with local files. See the documentation for details on the yaml.

PR #260, opened 2022-10-13
Update to xclim 0.38

aulemahal:
This branch is based on extend-dataset.

## Overview

This PR updates finch to use xclim 0.38.

Changes:

No changes were needed in the processes : all changes in finch come from changes in xclim.

For the documentation, the recent addition of a single reference bibfile in xclim made some acrobatics necessary. To correctly render the docstrings we parse from xclim, meaning having correct references, we need to have this reference file. The doc's conf downloads it each time it is run, using the installed xclim version to construct its github url.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -19,7 +19,10 @@
 # absolute, like shown here.
 #
 import os
+from pathlib import Path
 import sys
+from urllib.request import urlopen
+from xclim import __version__ as xcver
 
 # Add finch to sys.path to avoid having to full
 # install finch for autodoc.
@@ -46,6 +49,7 @@ extensions = [
     "sphinx.ext.imgconverter",
     "nbsphinx",
     "IPython.sphinxext.ipython_console_highlighting",
+    "sphinxcontrib.bibtex"
 ]
 
 # To avoid having to install these and burst memory limit on ReadTheDocs.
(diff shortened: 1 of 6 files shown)


docs/source/conf.py
```diff
@@ -101,7 +112,7 @@
 
 # General information about the project.
 project = "Finch"
-copyright = "2018-2020, David Huard"
+copyright = "2018-2022, David Huard"
```
Zeitsperre, line 115: Let's get future-proof:
```suggestion
copyright = f"2018-{date.today().year()}, David Huard"
```

docs/source/conf.py
```diff
@@ -46,6 +49,7 @@
     "sphinx.ext.imgconverter",
     "nbsphinx",
     "IPython.sphinxext.ipython_console_highlighting",
+    "sphinxcontrib.bibtex"
```
Zeitsperre, line 52: Don't forget to add this to the requirements!

Zeitsperre (approved):

aulemahal:
@perronld 

With this change (thus including #258), the list of available processes is almost not modified. We only have the following additions: cffwis, df, ffdi, kbdi, spei, spi.

The underlysing xclim indicators all use options and/or variables that make them incompatible with the ensemble processes.

As such, I am merging, this change was finally very small.

PR #264, opened 2022-11-08
Fixed iter_local to avoid files duplication in subset operations

perronld:
## Overview

This PR fixes files duplication on subset operations

Changes:

* Fixed iter_local when depth > 0 to avoid all files to be considered twice

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Changes
 *******
 
+TBD (unreleased)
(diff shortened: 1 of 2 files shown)


aulemahal:
The ci problem looks like a new behaviour for the new xarray, maybe? It came out 4 days ago.

Zeitsperre:
The fix should be in #259

PR #265, opened 2022-11-24
Add workflow cancellation action

Zeitsperre:
## Overview

Changes:

* Added an action for cancelling a previous CI workflow if/when two commits are pushed in rapid succession


huard (approved):

huard:
This only applies to code linting ?

Zeitsperre:
The code linting is the very first action in the `main` GitHub workflow, so if two commits both trigger workflows, the linting job triggered by the later commit will cancel the previous workflow entirely.

PR #272, opened 2023-01-27
Update xclim - Ensembles without percentiles - Clean up dev config handling

### LINKED ISSUE #270: Allow ensemble processes to return the dataset with individual realizations
## Description
For climatedata.ca users have expressed interest of being able to calculate indicators for individual model runs.  This could potentially be done by calling a large number individual finch processes but in my mind the simplest method forward would be to add a `calculate_percentiles` flag (default = True) to the ensemble_utils calls here : https://github.com/bird-house/finch/blob/38b3e322c140c22a6795e686f62809daafe9dd7c/finch/processes/ensemble_utils.py#L346

ideally we could also make use of the `create_ensemble()`  realizations parameter to add simulation ids as the realizations coordinate https://xclim.readthedocs.io/en/stable/xclim.ensembles.html#xclim.ensembles._base.create_ensemble

aulemahal:
## Overview

This PR fixes #270 
Changes:

* Update to xclim 0.40.0.
    - `fit` and  `stats`  are now `discharge_distribution_fit` and `discharge_stats`.
    - The two above and `freq_analysis` now take `discharge` as input (not `da`).

* Passing an empty string to `ensemble_percentiles` in ensemble processes will return the merged un-reduced ensemble. The different members are listed along the `realization` coordinates through raw names allowing for basic distinction between the input members.
* No more configuration handled by cli arguments to `finch start`. Instead of having 2 default config (`default.cfg` and those arguments), everything is handled by `default.cfg`. This removes the need for the templates folder and `jinja2`.
* No more replacement of WPS urls in the notebooks. Before, we were transforming the `localhost:5000` urls to `pavics.ouranos.ca` but changes in the folder hierarchy on the production server have made this method a bit awkward. 

## Related Issue / Discussion
https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/116

## Additional Information
The easy solution for populating the `realization` coordinate is to use the filenames. Those are constructed from the source, so there might be much more information than necessary. Also, for the multi-scenario case, I had to explicitly remove the "rcpXX" string from the name to allow correct concatenation.

A cleaner solution would be to propagate what our `Dataset` class has parsed from the filename, but that would demand either clean attributes all along or to pass filenames along the computation in a more structured manner.


tlogan2000 (approved):

tlogan2000:
@aulemahal can this be merged in your opinion?

aulemahal:
@tlvu The last CI was ran with the notebook sanitization from my PR, so it is working here at least!


@huard While trying to figure out how to fix the urlpath issue, I realized that  `finch main` was setting a default config (through default `click` args being written with a jinja template to `.custom.cfg`). This config was overriding the one in `default.cfg`. But the latter is more complete than the former. Thus, I decided to just remove this `.custom.cfg` method. Default configuration now _only_ relies on `default.cfg`.   However, my final solution to the notebook output issue (if we choose this one) removes the need I had to edit the default configuration. So, I don't mind putting the previous method back if it is needed in some cases I haven't considered. Do you have an opinion?

huard:
I think there are too many places where the config can be changed, and it gets confusing. If you removed one place, and things still work, I won't complain.

huard (approved):

huard:
owslib removed the verbose argument in https://github.com/geopython/OWSLib/pull/864
We'll need to release a new version of birdy.

huard:
I just released birdy 0.8.2, it should fix the test suite failure here.

PR #280, opened 2023-06-02
Add humidex calc

### LINKED ISSUE #278: Implement Humidex days over thresh calculation
## Description

In relation to climatedata.ca analysis page `ensemble` calculations we would like to implement a daily max humidity >= to threshold calculation

This should be able to recycle `atmos.tx_days_above` while having the input variable mapped as  `HXmax` instead of `tasmax`

Data source model list would be as in : https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/eccc/CCCS_humidex/Humidex/daily/catalog.html

tlogan2000:
## Overview

This PR fixes #278

Changes:
* Added hxmax_days_above calculation via yml module
* More globally finch can now support custom yml modules;  
* Exclusion / inclusion of subset of custom modules can be achieved via the config
* Automatic detection of associated new input variables (e.g. `HXmax` in this case)

## Additional Information
Adding french metadata to the humidex yml could be useful


tlogan2000:
Local tests are failing but might be due to updated xclim to `0.43`

```
ImportError while loading conftest '/home/logan/PycharmProjects/finch/tests/conftest.py'.
tests/conftest.py:14: in <module>
    from xclim.testing.tests.conftest import pr_hr_series, pr_series, q_series, tas_series
E   ModuleNotFoundError: No module named 'xclim.testing.tests'

```

requirements_docs.txt
```diff
@@ -2,7 +2,7 @@ pywps>=4.5.1
 sphinx>=4.0
 nbsphinx
 ipython
-xclim==0.38
+xclim==0.43
 pint<0.20  # xclim < 0.39 breaks with pint 0.20
```
Zeitsperre, line 6: Ditto.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 cftime
 pywps>=4.5.1
-xclim==0.38
+xclim==0.43
 pint<=0.19.2  # xclim < 0.39 breaks with pint 0.20
```
Zeitsperre, line 4: This pin is no longer needed, I imagine.

Zeitsperre: This should also be addressed in the environment files

aulemahal:
Yeah indeed... You'll need to translate all those "*_series" to calls  of https://xclim.readthedocs.io/en/v0.43.0/xclim.testing.html#xclim.testing.helpers.test_timeseries.

You might get inspiration from xscen's brand new test modules : https://github.com/Ouranosinc/xscen/blob/9b070c79762de28007f034743c25fffe52904900/tests/test_aggregate.py

environment.yml
Zeitsperre, line 1: You might want to put an upper pin on `python` here. Python3.11 is not officially support with some libraries.

requirements_docs.txt
tlogan2000: @aulemahal indicated that xclim is always pinned in finch? 
Also if I wait for #272 then integrate changes will the tests fix themselves?

Zeitsperre: I'm talking specifically the `pint` pin, sorry!

aulemahal: Merging should fix failures related to h5netcdf and the subset notebook, but not the ones from the "*_series" test helpers, these come from a change posterior to 0.40 (the xclim in 272).

finch/default.cfg
```diff
@@ -18,7 +18,8 @@ language=en-US,fr
 [finch]
 subset_threads = 1
 datasets_config = datasets.yml
-default_dataset = candcs-u5
+default_dataset = candcs-u6
+xclim_modules = humidex;streamflow
```
aulemahal, line 22: Hum, it seems the ";" can be used as a comment marker in the config (https://docs.python.org/3/library/configparser.html?highlight=configuration#supported-ini-file-structure).

May be we should use "," instead.

finch/processes/modules/humidex.yml
```diff
@@ -0,0 +1,29 @@
+module: humidex
+doc:
+realm: atmos
+indicators:
+  hxmax_days_above:
+    cf_attrs:
+      title: Number of days with maximum humidex above a given threshold
+      cell_methods: 'time: sum over days'
+      long_name: The number of days with maximum humidex above {threshold}
+      standard_name: number_of_days_with_humidex_above_threshold
```
aulemahal, line 10: This standard name doesn't exist!
```suggestion
```

finch/processes/utils.py
```diff
@@ -62,10 +63,28 @@
...
 
+
+def get_virtual_modules():
+    """Load virtual modules."""
+    modules = {}
+    if modfiles := get_config_value("finch", "xclim_modules"):
+        modfolder = [
+            f for f in Path(".").rglob("*") if f.is_dir() and f.name == "modules"
+        ][0]
+        for modfile in modfiles.split(";"):
```
aulemahal, line 74:
```suggestion
        for modfile in modfiles.split(","):
```
See above.

finch/processes/utils.py
```diff
@@ -62,10 +63,28 @@
...
+            modules[modfile] = dict(indicators=indicators)
+    return modules
+
+
 # Some other constants
-xclim_variables = set(
+xclim_variables = list(
     list_input_variables(submodules=["atmos", "land", "seaIce"]).keys()
 )
+xclim_variables = set(xclim_variables)
```
aulemahal, line 87: Is this still needed?

finch/processes/wps_xclim_indices.py
```diff
@@ -37,6 +36,7 @@ class XclimIndicatorBase(FinchProcess):
     """
 
     xci = None
+    allvars = []
```
aulemahal, line 39: Hum, I'm not sure, but having a mutable default on the class, outside the  `__init__` seems weird.
```suggestion
```

finch/processes/wps_xclim_indices.py
```diff
@@ -47,7 +47,10 @@ def __init__(self):
 
         outputs = [wpsio.output_netcdf_zip, wpsio.output_log, wpsio.output_metalink]
 
-        inputs = convert_xclim_inputs_to_pywps(self.xci.parameters, self.xci.identifier)
+        inputs, varnames = convert_xclim_inputs_to_pywps(
+            self.xci.parameters, self.xci.identifier
+        )
+        self.allvars.extend(varnames)
```
aulemahal, line 53:
```suggestion
        self.allvars = varnames
```
See above.

aulemahal (changes requested):
Seems good overall! 

A few changes requested.

This is also missing some documentation, principally in `docs/source/configuration.rst`.

finch/processes/utils.py
tlogan2000: wasn't sure .will try without it

finch/processes/utils.py
```diff
@@ -62,10 +63,28 @@
...
+
+def get_virtual_modules():
+    """Load virtual modules."""
+    modules = {}
+    if modfiles := get_config_value("finch", "xclim_modules"):
+        modfolder = [
+            f for f in Path(".").rglob("*") if f.is_dir() and f.name == "modules"
+        ][0]
+        for modfile in modfiles.split(";"):
+            mod = build_indicator_module_from_yaml(modfolder.joinpath(modfile))
```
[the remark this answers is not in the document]
tlogan2000: indeed. fogot I did this...  was having trouble with relative paths and just wanted to get it working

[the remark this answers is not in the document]
tlogan2000: Is this closer to what you would have in mind?
````
    modules = {}
    if modfiles := get_config_value("finch", "xclim_modules"):
        for modfile in modfiles.split(","):
            if os.path.isabs(modfile):
                mod = build_indicator_module_from_yaml(Path(modfile))
            else:
                mod = build_indicator_module_from_yaml(Path("finch").joinpath(modfile))
            indicators = []
            for indname, ind in mod.iter_indicators():
                indicators.append(ind.get_instance())
            modules[Path(modfile).name] = dict(indicators=indicators)
    return modules
````

Otherwise the config could map `finch/processes/modules/humidex;finch/processes/modules/streamflow`  with no need for an `if` statement

finch/processes/utils.py
tlogan2000: removed test pass locally

finch/processes/utils.py
tlogan2000: done

finch/processes/utils.py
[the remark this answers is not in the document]
tlogan2000: have implemented as above for now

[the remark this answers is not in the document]
aulemahal: Yes, that was what I had in mind!

However, instead of `Path("finch")`, maybe: `Path(__file__).parent.parent` ? The former depends on the current working directory.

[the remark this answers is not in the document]
tlogan2000: done

tlogan2000:
> Seems good overall!
> 
> A few changes requested.
> 
> This is also missing some documentation, principally in `docs/source/configuration.rst`.

Added some info to the configuration.rst   let me know if it seems ok

aulemahal (approved):

PR #291, opened 2023-07-10
Add dockerfile testing workflow

Zeitsperre:
## Overview

Changes:

* Added a dockerfile testing workflow
* Updated metadata in docker image
* Set an upper pin on Python (below v3.12)

## Related Issue / Discussion

https://github.com/Ouranosinc/raven/pull/481

### DIFF
--- a/.github/workflows/docker-testing.yml
+++ b/.github/workflows/docker-testing.yml
@@ -27,4 +27,7 @@ jobs:
           image: localfinch:latest
           options: -p 5000:5000
           run: |
-            gunicorn --bind=0.0.0.0:5000 -t 60 finch.wsgi:application
+            finch start -d
+            sleep 2s
+            finch status
+            finch stop
--- a/.github/workflows/docker-testing.yml
+++ b/.github/workflows/docker-testing.yml
@@ -8,9 +8,13 @@ on:
 
 jobs:
   build:
-    name: Build Docker image
+    name: Build and Test Docker image
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel previous runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+            access_token: ${{ github.token }}
       - uses: actions/checkout@v3
       - uses: docker/setup-buildx-action@v2
       - uses: docker/build-push-action@v4
--- a/.github/workflows/docker-testing.yml
+++ b/.github/workflows/docker-testing.yml
@@ -1,0 +1,30 @@
+name: Docker-based Testing Suite
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  build:
+    name: Build Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: docker/setup-buildx-action@v2
+      - uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: "Dockerfile"
+          tags: localfinch:latest
+          load: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          push: false
+      - uses: addnab/docker-run-action@v3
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -1,19 +1,25 @@
 # vim:set ft=dockerfile:
 FROM condaforge/mambaforge
-MAINTAINER https://github.com/bird-house/finch
+ARG DEBIAN_FRONTEND=noninteractive
+ENV PIP_ROOT_USER_ACTION=ignore
+LABEL org.opencontainers.image.authors="bourgault.pascal@ouranos.ca"
```
tlvu, line 5: Why not leave https://github.com/bird-house/finch as maintainer?  Not sure Pascal will want to receive all bugs report or requests.

Dockerfile
```diff
@@ -1,19 +1,25 @@
...
 WORKDIR /code
+COPY . ./
 
+# Build finch environment
 COPY environment.yml .
-RUN mamba env create -n finch -f environment.yml \
-    && mamba install -n finch gunicorn psycopg2 \
-    && mamba clean -ay
-
-COPY . .
```
tlvu, line 13: Keep this `COPY` below all the mamba install for layer re-use.  In a Dockerfile, we want the most frequently changed at the bottom for layer re-use to speed up and minimize space use during dev builds.

Dockerfile
```diff
@@ -1,19 +1,25 @@
...
 
+# Switch to /code directory
 WORKDIR /code
+COPY . ./
 
+# Build finch environment
 COPY environment.yml .
-RUN mamba env create -n finch -f environment.yml \
-    && mamba install -n finch gunicorn psycopg2 \
-    && mamba clean -ay
```
tlvu, line 11: Keep all install steps in the same `RUN` layer for proper clean up at the end.  If you perform in different layers, then the final clean up is useless since it can not clean up in other layers that itself.

Dockerfile
```diff
@@ -1,19 +1,25 @@
...
-COPY . .
+RUN mamba env create -n finch -f environment.yml
+RUN mamba install -n finch gunicorn psycopg2
+RUN mamba clean --all --yes
 
+# Add the finch conda environment to the path
 ENV PATH /opt/conda/envs/finch/bin:$PATH
 
-EXPOSE 5000
+RUN pip install . --no-deps
```
tlvu, line 21: Oh interesting.  I never notice the was no pip install for Finch !!!  This is harmless but does it brings anything more other than adding another layer and consuming a bit more spaces?

.github/workflows/docker-testing.yml
```diff
@@ -0,0 +1,33 @@
...
+          cache-to: type=gha,mode=max
+          push: false
+      - uses: addnab/docker-run-action@v3
+        with:
+          image: localfinch:latest
+          options: -p 5000:5000
+          run: |
+            finch start -d
+            sleep 2s
+            finch status
```
tlvu, line 36: This is really nice !!!

tlvu (changes requested):
Awesome docker testing workflow, thanks a lot.  Will help for prod for sure.

Some minor changes requested.

Dockerfile
Zeitsperre: I don't believe it adds much to the resources (I specified --no-deps to ensure we don't pull PyPI deps). It does offer the ability to `RUN` the service using `$ finch`, so that's nice.

.github/workflows/docker-testing.yml
Zeitsperre: Thanks! Glad to hear it!

Dockerfile
tlvu: right ok, keep this

tlvu (approved):
LGTM, minor comment for readability

PR #292, opened 2023-07-17
Prepare v0.11.2 and automate the release process

Zeitsperre:
## Overview

Changes:

* Prepared release v0.11.2.
* Added a GitHub Workflow to automatically bump the version number, and another to create tags from the version bumping process.
* Added a pre-commit hook for validating the ReadTheDocs configuration and GitHub Workflows.


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,63 @@
...
+      - docs/Makefile
+      - etc/**.cfg
+      - environment-docs.yml
+      - environment.yml
+      - finch/__version__.py
+      - setup.cfg
+      - setup.py
+
+jobs:
+  bump_patch_version:
```
tlvu, line 34: So I guess this bump patch version only, from the job name.  What if we want to bump the minor or major?  We have a way to tell the workflow to not perform automatically the patch bump so we can manually perform minor or major bump?

.github/workflows/tag-on-release.yml
```diff
@@ -0,0 +1,54 @@
...
+            steps.filter.outputs.cruft &&
+            steps.filter.outputs.changes &&
+            steps.filter.outputs.Dockerfile &&
+            steps.filter.outputs.conf &&
+            steps.filter.outputs.version
+        run: |
+          git config --local user.email "bumpversion[bot]@ouranos.ca"
+          git config --local user.name "bumpversion[bot]"
+          VERSION="$(grep -E '__version__' finch/__version__.py | cut -d ' ' -f3)"
+          git tag "${VERSION}"
```
tlvu, line 47: Careful, the tag has a `v` in front of the recorded version.

CHANGES.rst
```diff
@@ -1,14 +1,16 @@
 Changes
 *******
 
-0.12.0 (unreleased)
+0.11.2 (2023-07-17)
```
tlvu, line 4: Wrong date.  This should be updated only on the day this PR is merged.

tlvu: Huh, we went from `0.11.0` to `0.11.2` in the change log?

tlvu (changes requested):
A few errors spotted.  Otherwise, this will be greatly useful to avoid redundant manual interventions.

CHANGES.rst
Zeitsperre: We already tagged a version as `0.11.1`, but it has no release notes.

.github/workflows/tag-on-release.yml
Zeitsperre: Good catch!

.github/workflows/bump-version.yml
Zeitsperre: My understanding would be that we would perform the changes for a major/minor release within a PR, bumping the version ourselves.

This PR will always touch 5 files:
  - .cruft.json
  - CHANGES.rst
  - Dockerfile
  - docs/source/conf.py
  - finch/__version__.py

Changes to those files will not trigger a bump version, but changes to all those files will trigger the tagging action.

tlvu: So if we do it manually, we will also change all those files, which would trigger the tagging action again?

CHANGES.rst
tlvu: Release date for 0.11.2 still wrong?  Unless the tag already pushed before this PR merged?

Zeitsperre: I figured that I'll update the date of release just before we merge.,

.github/workflows/bump-version.yml
Zeitsperre: If we only trigger on push to "master", the action will not trigger on changes within a branch. I think the most important thing to take note of is that a "Prepare release" PR should not touch code.

_But what if it does?_

This is quickly becoming complicated. I may need to rethink this approach.

Zeitsperre:
I've made it so that bumping version and releasing is only done via workflow dispatch. Fewer moving parts. Much simpler.

tlvu:
> I've made it so that bumping version and releasing is only done via workflow dispatch. Fewer moving parts. Much simpler.

What is workflow dispatch?

huard (approved):
Please add a note in the developer docs about this.

Zeitsperre:
@tlvu I just need you to approve this before I can go forward. In case I don't see it today, would you mind merging when ready?

tlvu (approved):

tlvu:
@Zeitsperre merged the PR so the release date matches the one in `CHANGES.md`.  However we should update that date at the same time we perform `bump2version`.   Can be done in another PR.  Sample code: https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/.bumpversion.cfg#L11-L22

tlvu:
@Zeitsperre I guess for this time only I have to manually push the tag `v0.11.2`?

Zeitsperre:
Good idea for the date management. I didn't realize that we could leverage that functionality. Would simplify things in many projects.

The Action is already working: https://github.com/bird-house/finch/actions/workflows/bump-version-tag.yml (though, there was a bug in the naming code; fixed on master now).

We don't need to do any manual pushing of tags; I've added a PAT for that action specifically for this purpose.

PR #297, opened 2023-08-22
Package and Metadata Cleanup

Zeitsperre:
## Overview

Changes:

* Updated ReadTheDocs to use the new mambaforge version
* Addressed calls in GitHub Action usage that were emitting warnings
* Updated MANIFEST.in to include and exclude the relevant files for the source distribution
* Fixed the setup.py to only include the neccessary files in the wheel 
* Updated AUTHORS.rst to list more contributors
* Removed namespace file (`__init__.py`) from tests to ensure that they aren't treated like an importable package
* Updated pre-commit hooks
* Sorted requirements for legibility
* Removed Travis-CI shell script

## Additional Information

The next step will be to make this package PEP 517 and PEP 621 compliant. With the package in a better shape, converting the setup files should be easy.


aulemahal (approved):
Thanks! This looks like a good cleanup.

PR #315, opened 2023-10-31
use pivot table to create csv realization columns

### LINKED ISSUE #311: Distinct models should have distinct columns in CSV output
## Description

When executing ensemble process with an empty string for "ensemble_percentiles", if the requested output is CSV, individual model should have their own columns, instead of a single column for the result, and an additional column for realization. This will make the resulting CSV file directly useable for spreadsheet and graphing.

## Environment

* Finch version used, if any: commit 546b0e600aad56112129f98dac781b37817511a9

Links to other issues or sources: https://ccdpwiki.atlassian.net/browse/CLIM-380

tlogan2000:
## Overview

This PR fixes #311 

Changes:

* Use pandas `df.pivot` to create columns from `realization` index level

### DIFF
--- a/finch/processes/utils.py
+++ b/finch/processes/utils.py
@@ -657,8 +657,27 @@ def dataset_to_dataframe(ds: xr.Dataset) -> pd.DataFrame:
         ds["time"] = [y.replace(hour=12) for y in time_values]
         ds.time.attrs = attrs
     df = ds.to_dataframe().reset_index()
-    new_cols = [ll for ll in ["lat", "lon", "time"] if ll in df.columns]
-    df = df.sort_values(new_cols).set_index(new_cols)
+
+    if "realization" not in ds.dims:
+        new_cols = [ll for ll in ["lat", "lon", "time"] if ll in df.columns]
+    else:
+        new_cols = [
+            ll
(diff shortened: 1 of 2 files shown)


finch/processes/utils.py
```diff
@@ -657,8 +657,27 @@ def dataset_to_dataframe(ds: xr.Dataset) -> pd.DataFrame:
...
     df = ds.to_dataframe().reset_index()
-    new_cols = [ll for ll in ["lat", "lon", "time"] if ll in df.columns]
-    df = df.sort_values(new_cols).set_index(new_cols)
+
+    if "realization" not in ds.dims:
+        new_cols = [ll for ll in ["lat", "lon", "time"] if ll in df.columns]
+    else:
+        new_cols = [
+            ll
+            for ll in ["lat", "lon", "time", "realization", "scenario", "region"]
```
aulemahal, line 666: Can't we just remove `realization` from this list instead of removing it below twice ?

aulemahal (approved):
Looks good!

A nitpicky comment, but that's all.

finch/processes/utils.py
tlogan2000: done

PR #322, opened 2023-12-04
Fix cftime bug, no fail on pip check, pin xarray/pandas

Zeitsperre:
## Overview

Changes:

* Added pins to xarray and pandas to prevent issues with breaking changes.
* Fixed an error with a cftime index fixing function.
* Switched `bump2version` for `bump-my-version`
* Do not fail when running pip check in conda environments

## Related Issue / Discussion

https://github.com/Ouranosinc/xclim/issues/1534
https://github.com/pydata/xarray/issues/8394


tlogan2000 (approved):

Zeitsperre:
This needs a pin on xarray to deal with the cftime issue.

PR #325, opened 2023-12-06
Use .bumpversion.toml configuration

Zeitsperre:
## Overview

Changes:

* Moves bump-my-version to TOML standard

## Related Issue / Discussion

This will be part of the eventual migration to the new packaging standard (`pyproject.toml`)


Zeitsperre:
@huard

I'll be tagging a new version after this. I think we can throw this onto PyPI afterwards!

aulemahal (approved):

PR #351, opened 2024-03-19
Docs and pip packaging

Zeitsperre:
## Overview

Changes:

* Slightly reorganized the documentation for easier navigation
* Fast-forwarded the cookiecutter
* Fixed the Manifest to add all necessary files to wheel
* Removed references to files that have never existed (`apidoc`)
* Cleaned up the setup code
* Added more files to be ignored in the `.gitignore` and in the `Manifest.in`

Breaking Changes:

* Renamed the package from `finch` to `birdhouse-finch`

This is needed in order to index `finch` on PyPI. This isn't the first time that there is a naming situation (`birdhouse-birdy`), so I don't imagine this is too controversial. The docker deployment and `Makefile` run services are not affected by this change.

* The documentation layout has been revised

Many directions were duplicated, and it was easy to become confused as to what was possible from the command line. This has been revised by moving the commands to run `finch` to `deploying`. I've also added some more directions on how best to install `finch`.

## Related Issue / Discussion

I will be performing a release on GitHub **and** on PyPI after this. I can confirm that the packaging works from both wheel and tarball.


aulemahal (approved):
Looks good to me!

aulemahal:
I think the toc hierarchy is wrong : 
image

Indicator processes are on the same level as the title "indicator processes".

Zeitsperre:
OK, I think this is good to go.

Docs nesting was a major problem since the virtual module is recognized differently than a true module, so there isn't any hierarchy in its members. This is dealt with now.

Asking for a re-review as changes has drifted into the rest of the documentation as well.

aulemahal (approved):
CDLABPM (Ça d'l'air bon pour moi)!

docs/source/deploying.rst
```diff
@@ -0,0 +1,100 @@
...
+
+   $ finch start --daemon  # start service as daemon
+   loading configuration
+   forked process id: 42
+
+The deployed WPS service is available by default on port 5000:
+
+http://localhost:5000/wps?service=WPS&version=1.0.0&request=GetCapabilities.
+
+.. note:: Remember the process ID (PID) so you can stop the service with ``kill PID``.
```
aulemahal, line 26: I know you only copied this, but doesn't  `finch` poop a `pywps.pid` file so that we don't have to remember a pid ?

Zeitsperre: It does! Maybe I should make reference to it instead. Noted.

CHANGES.rst
```diff
@@ -4,6 +4,8 @@ Changes
 0.11.5 (unreleased)
 ===================
 * Added a Makefile recipe to the repository to evaluate notebooks while ignoring the output cells.
+* Cleaned up documentation to facilitate easier navigation.
+* Fixed a few notebooks that were failing Jenkins build tests.
```
tlvu, line 8: What Finch notebook is failing the Jenkins build?  Nothing is failing on Jenkins since March 17 with regards to Finch notebooks.

tlvu (approved):
LGTM.  Don't see any notebook change in this PR.  The changelog needs to be fixed.

CHANGES.rst
tlvu: Opps, does this changelog includes other PR between now and the previous release? If yes then keep the note about notebook update.

PR #352, opened 2024-03-25
Prepare v0.12.0, conditional bump version

Zeitsperre:
## Overview

Changes:

* Updated the release version string (v0.12.0) and date (today)
* Updated versioning to use SemVer v2.0.0
* Update bump-version action to bump when changes are made to specific files on the master branch

## Related Issue / Discussion

The version scheme and bumping setup is the same as can be found for xclim and other Ouranos projects.

## Additional Information

https://semver.org/spec/v2.0.0.html


.github/workflows/bump-version-tag.yml
```diff
@@ -37,24 +59,33 @@ jobs:
...
+          python -m pip install bump-my-version>=0.18.3
+      - name: Conditional Bump
+        id: bump
+        run: |
+          if [[ ${{ env.CURRENT_VERSION }} =~ -dev(\.\d+)? ]]; then
+            echo "Development version (ends in 'dev(\.\d+)?'), bumping 'build' version"
+            bump-my-version bump build
+          else
+            echo "Version is stable, bumping 'patch' version"
+            bump-my-version bump patch
```
tlvu, line 81: Version bump automatically on each merge to `master`?  That's agressive !

.github/workflows/bump-version-tag.yml
```diff
@@ -1,28 +1,50 @@
...
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - .*
+      - .github/*/*.md
+      - .github/*/*.yml
+      - AUTHORS.rst
+      - CHANGES.rst
+      - Dockerfile
```
tlvu, line 13: This `paths-ignore` section is to not bump if those files listed here change?  Why would we not bump and make a release if `Dockerfile` changes?  Same for `environment.yml`.  Those files are part of the release docker build.

tlvu (approved):
I agree with the spirit of the PR but have a few questions.

.github/workflows/bump-version-tag.yml
Zeitsperre: It bumps the version to `vX.Y.Z-dev.#` to indicate that the master branch is different from the last release or the last PR merge. We don't release until we perform a `bump-my-version bump release` to drop the `-dev`.

.github/workflows/bump-version-tag.yml
Zeitsperre: I can remove the `environment.yml` files, no problem there.

The reason for not including the `Dockerfile` is because the version string is changed in the `Dockerfile` on `bump-my-version`, which would trigger `bump-my-version` to augment the version, which would cause changes in the `Dockerfile`, which would trigger `bump-my-version` to augment the version, which would cause changes in the `Dockerfile`...

.github/workflows/bump-version-tag.yml
tlvu: Ah ok, so each merge do not auto create a release, good then.

.github/workflows/bump-version-tag.yml
tlvu: I see, to avoid infinite loop. Okay, got that, good with leaving `Dockerfile` out.

PR #373, opened 2024-06-25
Prepare v0.12.1

### LINKED ISSUE #358: `unidecode` is under GPLv2
## Description

`unidecode` is used in several processes and, unfortunately, is not technically allowed to be integrated into this project due to incompatible licenses. This library needs to be removed and replaced.

## Additional Information

https://www.apache.org/licenses/GPL-compatibility.html

Zeitsperre:
## Overview

This PR fixes #358

Changes:

* Synchronized dependencies
* Prepared the next patch release (v0.12.1)
* Added a workaround to deal with changes in behaviour with pandas v2.0.0

## Discussion

It looks like `pandas` v2.0 changed some of the default behaviour with the casting of certain data types. For the `geoseries` data, the `LOCAL_DATE` dimension is not recognized as the `time` dimension, nor is it recognized as `datetime64[ns]` (instead it is `datetime64[ms]`, which is less well handled.).

This might be fixed in later versions of `pandas`.

Relevant issue: https://github.com/pydata/xarray/issues/7197


aulemahal (approved):
Looks good!

PR #427, opened 2024-10-31
Use xscen new

tlogan2000:
## Overview

This PR fixes [issue id]

Changes:

* Added ...

## Additional Information

### DIFF
--- a/.github/workflows/bump-version-tag.yml
+++ b/.github/workflows/bump-version-tag.yml
@@ -38,7 +38,7 @@ jobs:
       actions: read
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
+        uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
         with:
           disable-sudo: true
           egress-policy: block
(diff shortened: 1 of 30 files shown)


tlogan2000:
Local run of test suite are struggling triggering pull request to see test failures.

tlogan2000:
@aulemahal  @Zeitsperre  ... any idea why I get a permission error for `test_wps_ensemble.py::test_ensemble_hxmax_days_above_grid_point`  ??

aulemahal:
I really can't see why... All the other tests pass and they do the same thing...

aulemahal:
@tlogan2000 Hypothèse : the issue stems from temporary directories. I switched most (all?) temp file handling in the tests to use pytest's functions instead of custom ones. 

The notebook error were coming from a faulty `is_opendap_url` function. The function was performing a test that worked with Thredds v4, but does not with v5. From what I understand of the OpenDAP doc, the previous test was not *required* to work, but  my change should. 

In details : we were making a HEAD request on the "raw" URL. OpenDAP does not specify what the server should return when hitting the pure dataset url without any query. Thredds v4 was returning something, v5 is returning an error. My solution is to request the "metadata" of the dataset using the ".dds" "query", which is something all OpenDAP server should be able to do.

aulemahal:
Changing the temp dir methods didn't work, but dropping python 3.9 did. 

Any reasons not to,  @Zeitsperre ? I think finch is solely used by ClimateData and through it's docker image which installs python 3.12 by default I think.

Zeitsperre:
@aulemahal Dropping Python3.9 is fine. We've already done this in recent versions of `xclim` and `xscen`. This is a web service, so an especially wide range of compatible Python versions isn't as important.

Zeitsperre (approved):
LGTM

I think I might update the cookiecutter once this is merged in, then we can release a new version. Great work!

aulemahal (approved):

Zeitsperre:
This is just waiting on an update to the changelog. Thanks again!

PR #441, opened 2025-01-10
Update build system, update cookiecutter template, use vulture and ruff

Zeitsperre:
## Overview

Changes:

* Updated the cookiecutter template to the latest commit
* Enabled `ruff` and `vulture` in pre-commit
* Changed build system from `setuptools` to `flit`
* Removed several obsolete configuration files

## Related Issue / Discussion

The Code of Conduct is still missing a contact email (should we use the one for Ouranos projects?)

Some TODOs/FIXMEs related to security can be ignored for now, but a refresh/rebuild of `finch` will need to address these.

There are still two tests related to file access that seem to be failing for me locally. Not sure why. It's possible that they might just be due to older NetCDF4 libraries. A fresh environment seems to work fine.

## Additional Information

https://github.com/bird-house/cookiecutter-birdhouse
https://github.com/astral-sh/ruff
https://github.com/jendrikseipp/vulture


Zeitsperre:
The docs are raising warnings due to some issues with `xclim`, I imagine:
```
/home/docs/checkouts/readthedocs.org/user_builds/finch/conda/441/lib/python3.12/site-packages/finch/processes/wps_base.py:docstring of finch.processes.wps_base.discharge_stats_Indicator_Process:9: WARNING: undefined label: 'timeseries.resampling' [ref.ref]
/home/docs/checkouts/readthedocs.org/user_builds/finch/conda/441/lib/python3.12/site-packages/finch/processes/wps_base.py:docstring of finch.processes.wps_base.doy_qmax_Indicator_Process:7: WARNING: undefined label: 'timeseries.resampling' [ref.ref]
/home/docs/checkouts/readthedocs.org/user_builds/finch/conda/441/lib/python3.12/site-packages/finch/processes/wps_base.py:docstring of finch.processes.wps_base.doy_qmin_Indicator_Process:7: WARNING: undefined label: 'timeseries.resampling' [ref.ref]
/home/docs/checkouts/readthedocs.org/user_builds/finch/conda/441/lib/python3.12/site-packages/finch/processes/wps_base.py:docstring of finch.processes.wps_base.hxmax_days_above_Indicator_Process:9: WARNING: undefined label: 'timeseries.resampling' [ref.ref]
```

@aulemahal do you have any idea what that could be about? Has this perhaps been fixed in newer versions (v0.52.2 is installed here).

@tlogan2000 The only failing tests I can see are from the ensembles tests not returning datasets with the proper `time` dimension length. Not sure why that would be all of a sudden.

tlogan2000 (approved):
@aulemahal will leave a comment for the docs warnings (linking to pandas docs in our docstrings that is in xclim but not in the conf here). In terms of the code all looks okay

aulemahal:
Sorry  I forgot about your question!

The issue in the doc is because the docstring of the process contains the ``:ref:`timeseries.resampling` `` markup and that label is not defined anywhere in finch. It comes from xclim, the docstring is taken from the indicator which takes it from the indice. The label is not defined in xclim either, but xclim uses `intersphinx` and adds `pandas` to the config. In xclim, the reference is ultimately resolved as a Pandas page.


Example : Process [doy_qmax](https://pavics-sdi--441.org.readthedocs.build/projects/finch/en/441/processes/indicators.html#finch.processes.virtual.indicators.doy_qmax_Indicator_Process) has the markup in the doc for argument 
-`freq`. This text is taken (after many wraps and copies) from [select_resample_op](https://github.com/Ouranosinc/xclim/blob/a57288966eb27afff5b90c67b54afc34c2decc83/src/xclim/indices/generic.py#L72). 

One simple solution is to ignore.

One complex and unnecessary solution is to setup intersphinx and add pandas as a potential source, [as done in xclim](https://github.com/Ouranosinc/xclim/blob/a57288966eb27afff5b90c67b54afc34c2decc83/docs/conf.py#L180C3-L180C71). I say unnecessary because finch's documentation is mostly useless. The service is only used in ClimateData's backend : nobody ever sees the docstring and even when they do, it's in its raw, WPS-response format.

Zeitsperre:
@aulemahal Thanks for the explanation. That makes a lot of sense.

This should come as no surprise, but my tendency would be to add the missing `intersphinx` references. 
 
> One simple solution is to ignore.

The issue with ignoring the warnings (`fail_on_warning: false`) is that it also ignores whether large sections of the documentation fail to build at all. This was an issue with other `Birdhouse` services, as their processes pages were completely void of any documentation; All that mattered was whether the documentation built at all.

If this is a one-line fix for `intersphinx`, I don't see the harm in adding it. The documentation here is rendered and referenced from other `Birdhouse`/`DACCS`/etc. pages.

PR #491, opened 2025-05-12
Add time horizons

tlogan2000:
## Overview

This PR fixes [issue id]

Changes:

* add functionality to calculate 30 year mean on ensemble calculations


aulemahal (approved):
Looks good!

src/finch/processes/ensemble_utils.py
```diff
@@ -356,6 +356,7 @@ def make_ensemble(  # noqa: D103
     files: list[Path],
     percentiles: list[int],
     spatavg: bool | None = False,
+    tmpavg: bool | None = False,
```
aulemahal, line 359:
```suggestion
    tmpavg: bool  = False,
```
Nit pick, but I don't think there's a way to pass None here, and it wouldn't mean anything.

PR #537, opened 2026-05-21
Update to xclim 0.61, xsdba 0.6.1 and xscen 0.15

### LINKED ISSUE #657: :bug: [BUG]: Finch: Switched abstracts for `heat_wave_max_length`/`heat_wave_total_length` processes in GetCapabilities
## Summary

Finch `v0.12.1` pinned in birdhouse-deploy exposes swapped abstracts for the `heat_wave_max_length` and `heat_wave_total_length` processes in GetCapabilities. 

## Details

`birdhouse-deploy` pins finch `v0.12.1` via: https://github.com/bird-house/birdhouse-deploy/commit/565289e59847055352d50858972a4dc64e09237e, affected versions: 2.16.13-2.24.1

This is a downstream effect of a bug that has been fixed in xclim; see the corresponding [finch issue](https://github.com/bird-house/finch/issues/524) for details.

### To Reproduce

1. Deploy a birdhouse-deploy stack with finch pinned at `v0.12.1`
2. Issue a WPS GetCapabilities request to the running finch service
3. Inspect the abstract fields for the `heat_wave_max_length` and `heat_wave_total_length` processes

## Environment

| Information           | Value
| --------------------- | --------------------------------------------------------------
| Server/Platform URL   | https://garryoak.pacificclimate.org/
| Version Tag/Commit    | 2.20.1
| Related issues/PR     | https://github.com/bird-house/finch/issues/524
| Related components    | finch
| Custom configuration  | `birdhouse/components/finch/default.env`

## Concerned Organizations
This bug is verified for all organizations.

aulemahal:
## Overview

This PR:
- fixes bird-house/birdhouse-deploy#657
- fixes #524 

Changes:

* Unpins a lot of things
* Fixes a few issues to work with xclim 0.61
* Last commit fixes around 600 warnings, for future-proofing and log readability.

## Related Issue / Discussion
@fmigneault Could you check my changes about sentry ? This was for a warning about deprecated changes.

## Additional Information
If there are breaking changes, they should be related to the xclim changes, I don't think I introduced anything new ;).


src/finch/processes/__init__.py
```diff
@@ -1,6 +1,7 @@
 # noqa: D104
 import logging
 
+import xarray as xr
```
Zeitsperre: The `xarray` import doesn't seem to be referenced anywhere?

src/finch/processes/wpsio.py
```diff
@@ -256,7 +256,7 @@ def get_ensemble_inputs(novar=False):
     abstract="Method used to determine which aggregations should be considered missing.",
     data_type="string",
     default=OPTIONS[CHECK_MISSING],
-    allowed_values=list(MISSING_METHODS.keys()),
+    allowed_values=list(MISSING_METHODS.keys()) + ["skip"],
```
Zeitsperre: Can we use `append()` here? I just don't like math.

environment.yml
Zeitsperre, line 1: I think we need `xsdba` here.

pyproject.toml
Zeitsperre, line 1: `xsdba` is missing here.

pyproject.toml
```diff
@@ -28,6 +28,8 @@ classifiers = [
   "Programming Language :: Python :: 3.10",
```
Zeitsperre: Should we drop Python3.10?

environment-docs.yml
Zeitsperre, line 1: I think we need xsdba here.

environment-docs.yml
```diff
@@ -1,18 +1,17 @@
...
   - matplotlib-base >=3.5.0
-  - nbsphinx >=0.9.8
-  - pandas >=2.2.0,<3.0
+  - nbsphinx >=0.9.5
+  - pandas >=2.2.0
   - pywps >=4.5.1
-  - setuptools >=78.1.1
-  - sphinx >=8.2.0 # Pinned until nbsphinx supports Sphinx 8.2
+  - setuptools >=71.0.0
+  - sphinx >=7.0.0
```
Zeitsperre: If we drop Python 3.10, we can bump this to v9.0.

pyproject.toml
```diff
@@ -37,68 +39,65 @@ dependencies = [
...
-  "xarray >=2023.11.0,<2025.4.0",
-  "xclim ==0.52.2", # remember to match xclim version in environment.yml as well
-  "xesmf >=0.6.2,!=0.8.8",
-  "xscen ==0.10.0", # remember to match xscen version in environment.yml as well
-  "werkzeug>=3.0.6",
-  # Temporary fixes
-  "fastprogress <1.1" # FIXME: Temporary fix following https://github.com/intake/intake-esm/pull/773. Remove when intake-esm is updated in xscen.
+  "siphon",
+  "xarray >=2023.11.0",
+  "xclim ==0.61.0", # remember to match xclim version in environment.yml as well
```
Zeitsperre: I think we should be allowing patch releases here. We'll probably have two before v1.0:
```suggestion
  "xclim >=0.61.0,<0.62.0", # remember to match xclim version in environment.yml as well
```

environment.yml
```diff
@@ -1,36 +1,33 @@
 name: finch
 channels:
   - conda-forge
-  - nodefaults
 dependencies:
-  - python >=3.10,<3.13
-  - pip >=25.3.0
+  - python >=3.10,<3.14
+  - pip >=24.2.0
```
Zeitsperre: We should push this much higher.
```suggestion
  - pip >=26.1.0
```

Zeitsperre (changes requested):
I think we should be making a few breaking changes here, seeing as we're jumping several versions in xclim and xscen.

src/finch/processes/__init__.py
aulemahal: oups, debugging leftovers.

src/finch/processes/wpsio.py
aulemahal: done. `append` does it in-place, so it's more lines, but we don't have a quota ;)

pyproject.toml
aulemahal: Why not. Done!

environment-docs.yml
aulemahal: I didn't add it and made it a mockup instead, let's see what that does.

aulemahal: I removed the mockup stuff it was not being used anyway as pip was installing everything in the RTD env.

aulemahal:
@Zeitsperre haha I had some trouble with the CI...

Installing in hybrid : conda env + dev deps in pip was giving broken envs after my unpins. I rationalized it a bit like we do elsewhere : one single env file installing normal deps as well as dev and docs deps.  
I removed the `develop` and `install` entries of the Makefile to reflect this change of philosophy.

For ReadTheDocs, we had a custom funky setup that was created to avoid running out of memory on the remote build. This problem seems gone, here I simply install the full normal env and don't do any mockups.

aulemahal:
Ach, I just saw that this change the Docker image by installing all unneeded dev tools in it... May be we could use pip in the docker image ? Or maybe it's not a problem.

Zeitsperre:
I'll give this a proper once-over next week!

Makefile
```diff
@@ -38,17 +38,6 @@ help: ## print this help message. (Default)
...
-
-.PHONY: install
-install: ## install finch application
-	@echo "Installing application ..."
-	@-bash -c 'pip install -e .'
-	@echo "\nStart service with \`make start\` and stop with \`make stop\`."
-
-develop: ## install finch application with development libraries
-	@echo "Installing development requirements for tests and docs ..."
-	@-bash -c 'pip install -e ".[dev]"'
```
fmigneault, line 50: I don't understand why removing those? They should stay, as they would pick up the `pyproject.toml` automatically. I agree about removing the environment YAML files.

The idea to have them is that they create an abstraction of how it gets invoked, so you don't need to change the GitHub Workflow calls (or anywhere else that would need them as well).

pyproject.toml
```diff
@@ -12,7 +12,7 @@ maintainers = [
   {name = "Trevor James Smith", email = "smith.trevorj@ouranos.ca"}
 ]
 readme = {file = "README.rst", content-type = "text/x-rst"}
-requires-python = ">=3.10.0"
+requires-python = ">=3.11.0"
```
fmigneault, line 15: `requires-python = ">=3.11.0,<4"` just in case (or even `<3.15` if we want the same as the classifiers and the `environment.yaml`).

environment.yml
```diff
@@ -28,9 +28,39 @@ dependencies:
...
-  - xclim =0.52.2  # remember to match xclim version in requirements_docs.txt as well
+  - xarray >=2023.11.0
+  - xclim >=0.61,<0.62 # remember to match xclim version in pyproject.toml as well
   - xesmf >=0.8.2,!=0.8.8
-  - xscen =0.10.0  # remember to match xscen version in environment.yml as well
+  - xsdba =0.6.1   # remember to match xsdba version in pyproject.toml as well
+  - xscen =0.15.0  # remember to match xscen version in pyproject.toml as well
   # Temporary fixes
   - fastprogress <1.1  # FIXME: Temporary fix following https://github.com/intake/intake-esm/pull/773. Remove when intake-esm is updated in xscen.
+  # Dev
```
fmigneault, line 37: At this point, is there even any reason to have the `environment.yaml`? 
Why not just `conda create -n finch && pip install .` combo all from the same `pyproject.toml` ?  Any package that still depends on `conda-forge` build specifically?

environment-docs.yml
```diff
@@ -1,18 +0,0 @@
...
-  - anyascii >=0.3.0
-  - birdy >=0.8.1
-  - ipython >=8.5.0
-  - matplotlib-base >=3.5.0
-  - nbsphinx >=0.9.8
-  - pandas >=2.2.0,<3.0
-  - pywps >=4.5.1
-  - setuptools >=78.1.1
-  - sphinx >=8.2.0 # Pinned until nbsphinx supports Sphinx 8.2
-  - sphinxcontrib-bibtex >=2.6.0
```
fmigneault, line 16: Sphinx and related specifically for docs should be in a `docs` extra in `pyproject.toml`.

docs/source/dev_guide.rst
fmigneault, line 1: To edit according to other comments.

fmigneault (changes requested):

Makefile
Zeitsperre: I'm with Francis as well on this. I see the Makefile as a more convenient way of running things that rely on the `pyproject.toml` for CI operations. The cookiecutter this is based on should be using this more as well.

pyproject.toml
Zeitsperre: I wouldn't add that here, but I would constrain the `environment.yml` to `>=3.11.0,<3.15`. It is entirely possible that the changes in Python 3.15 will not be breaking for `finch`.

Makefile
aulemahal: Yep sorry I went overboard here. My thinking was that the hybrid conda+pip was not wanted anymore. But those are still valid for pip-only envs.

environment.yml
Zeitsperre: There are a few things in here that rely on GDAL. It's also much easier to recreate the environment based on `environment.yml`. I'd also use this file to constrain the Python version installed in the image as well.

The Anaconda devs are building a new tool for better support of PyPI packages in conda, but it is still too easy to install some libraries in a broken way when mixing conda and PyPI. Perhaps if that tool gets better support.

aulemahal: For the development, I prefer using conda.

Also, we have a few averaging processes that use ESMF under the hood and ESMF is C/Fortran, not available on pypi. It's much easier to install with conda. 

For the deployment, we could use your suggestion though, which would avoid installing too much dev deps in the production build.

environment-docs.yml
Zeitsperre: I think I'm going to adjust this upstream, using the new PEP 735 conventions. If Pascal wants to move forwards with this, I can readjust it after the fact.

aulemahal: There's already one!

fmigneault (approved):
Assuming the docs dependencies will be re-added in a subsequent PR, that looks fine.

aulemahal:
@fmigneault , not sure I understand, aren't the docs dependencies already here : https://github.com/bird-house/finch/blob/180d9d250f3ca0c65ab4de620b57d406e34a8cdb/pyproject.toml#L94-L105 ?

Also, on my VM, the compressed docker image is 1600 MB, which is more than the current 900 MB, for you information.

fmigneault:
@aulemahal 
My bad. I didn’t notice that commit. You can ignore my comment. All good.

.github/workflows/main.yml
```diff
@@ -69,14 +69,16 @@ jobs:
           cache-downloads: true
           cache-environment: true
           environment-file: environment.yml
+          environment-name: finch
           create-args: >-
             python=${{ matrix.python-version }}
       - name: Install finch-wps
         run: |
-          make develop
+          pip install --no-deps -e .
```
Zeitsperre: Verbosity is great
```suggestion
          python -m pip install --no-deps --editable .
```

.github/workflows/main.yml
```diff
@@ -69,14 +69,16 @@ jobs:
...
       - name: Check versions
         run: |
           python -m pip check
+          python -m pip list
```
Zeitsperre: We can glean more info from micromamba
```suggestion
          micromamba list
          python -m pip check
```

docs/source/installation.rst
```diff
@@ -39,12 +39,12 @@ Check out code from the Finch GitHub repo and start the installation:
    $ git clone https://github.com/bird-house/finch.git
    $ cd finch
 
-Create Conda environment named `finch`:
+Create Conda environment named `finch`, this includes the development and documentation dependencies :
```
Zeitsperre:
```suggestion
Create Conda environment named `finch` (including the development and documentation dependencies):
```

docs/source/installation.rst
```diff
@@ -54,17 +54,10 @@ Install `finch` app:
...
-  OR
-  $ make develop
 
 Install from Conda
 ------------------
 
 .. note::
 
-   `finch` is not yet available on conda-forge. But we are working on making this package available soon!
+   `finch` is not yet available on conda-forge. We are not working on making this package available soon.
```
Zeitsperre: Are you sure about that? https://pypi.org/project/birdhouse-finch/

src/finch/cli.py
```diff
@@ -301,3 +301,7 @@ def start(
     else:
         # no daemon
         _run(app, bind_host=bind_host)
+
+
+if __name__ == "__main__":
+    cli()
```
Zeitsperre: Was the CLI not working previously??

.readthedocs.yml
```diff
@@ -21,12 +21,10 @@ build:
...
-  environment: environment-docs.yml
+  environment: environment.yml
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
   install:
     - method: pip
       path: .
-      extra_requirements:
-        - docs
```
Zeitsperre, line 32: I'd keep this here anyway. If dependencies desynchronize, then we know on RTD.

environment.yml
```diff
@@ -28,9 +28,39 @@ dependencies:
...
   - siphon >=0.10.0
   - werkzeug >=3.0.6
-  - xarray >=2023.11.0,<2025.03.0
-  - xclim =0.52.2  # remember to match xclim version in requirements_docs.txt as well
+  - xarray >=2023.11.0
+  - xclim >=0.61,<0.62 # remember to match xclim version in pyproject.toml as well
   - xesmf >=0.8.2,!=0.8.8
-  - xscen =0.10.0  # remember to match xscen version in environment.yml as well
+  - xsdba =0.6.1   # remember to match xsdba version in pyproject.toml as well
+  - xscen =0.15.0  # remember to match xscen version in pyproject.toml as well
```
Zeitsperre: We've already put out a patch for 0.15. We shouldn't pin the patch version.
```suggestion
  - xscen >=0.15.0,<0.16.0  # remember to match xscen version in pyproject.toml as well
```

environment.yml
```diff
@@ -3,19 +3,19 @@ channels:
...
   - clisops >=0.16.2
-  - dask >=2023.5.1,<2025.3.0
+  - dask >=2023.5.1
   - distributed
   - geopandas >=1.0
   - jinja2 >=3.1.4
-  - netcdf4 <=1.7.2
-  - numcodecs <0.16.0
+  - netcdf4
+  - numcodecs
```
Zeitsperre: I think numcodecs is still pinned in xscen?

Zeitsperre (approved):
Looks pretty good. There are more than a few things that need to be updated, but I'll get to them in the template.

If there isn't any urgency, we can merge this, perform additional updates and release on PyPI. Most importantly, the docker image needs to build properly if we want to use it with PAVICS.

docs/source/installation.rst
aulemahal: Hmm, this is on pypi ? I don't think there's work in progress to put it on conda-forge ? There's also no need I think.

src/finch/cli.py
aulemahal: It was, but I had to test something without having the package installed, so I needed to call the file directly. I can remove this.

environment.yml
aulemahal: indeed. But I realize now that we never explicitly import `numcodecs`, so maybe we don't need it in the deps in the first place ? We require xscen which requires and pins numcodecs, seems enough to me!

.github/workflows/main.yml
fmigneault: I'm confused. You agreed that using `make` command was good for the abstraction, making the CI generic and reusable, but then want them explicitly? 

I don't understand why that is needed. If you want `make` to display the invoked command explicitly, just remove the `@` before the specific commands desired.

Zeitsperre: True, I prefer using `make`. Sorry!

Zeitsperre:
We're running into issues because of `pytest-notebooks`, which is quite out of date. We replaced this with `nbval` (although this library is very quickly becoming out of date as well).

.github/workflows/main.yml
aulemahal: @Zeitsperre `pip check` fails because `pytest-notebook` has a pinned dependency in pip's world that is not pinned in conda-forge. Now that we only use conda for the CI env, maybe we could remove the `pip check` call ? All-mamba is much less likely to create broken environments than the previous hybrid setup.

aulemahal:
Lol I think `pytest-notebook` is already unneeded,  but was not removed from the env. Simple fix!

.readthedocs.yml
aulemahal: done.

src/finch/cli.py
aulemahal: Debugging stuff removed.

.github/workflows/main.yml
aulemahal: pytest-notebook removed as already uneeded.

.github/workflows/main.yml
aulemahal: - make put back (this has the consequence of letting pip install dependencies as `--no-deps` is not in the make rule).
- make commands using the full explicit formulation

docs/source/installation.rst
Zeitsperre: True.

```suggestion
   There are no plans to make `finch` available on conda-forge.
```

PR #544, opened 2026-07-03
avoid duplicate copies of docker layers

fmigneault:
## Overview

Reworks the docker command to avoid duplicate copy of file contents.
Because of the introduced `chown nonroot`, that basically forces the docker layer to cache all installed package files twice!

```
birdhouse/finch:version-0.13.2                  289b1b21b679       2.58GB
birdhouse/finch:0.13.3-dev.3                    0e7982021ae3       4.85GB  <---
birdhouse/finch:0.13.3-dev.3-with-chown-copy    e94cb8840170       2.77GB
```

The update uses a preemptive user creation and then `COPY --chown` with that user to return close to the original size (assumed the diff is from other package artifacts).

I also attempted removing unnecessary dev/doc deps from `environment.yml` to reduce size (since it is used for installation in the docker).
However, many doc deps (`sphinx` and co.) remain. I suspect this is caused by https://github.com/geopython/pywps/blob/main/environment.yml which also includes dev/doc packages unnecessarily.

Normally (like in https://github.com/crim-ca/weaver/blob/307478adfb1a220cc57ced5e32c74ec6431f1679/requirements.txt#L122 does), a **pip** install of `pywps` will NOT include the dev/doc packages (I can confirm this in the Weaver docker image). However, because `environment.yml` is used in `finch` with conda-forge packages, I assume https://anaconda.org/channels/conda-forge/packages/pywps/overview is used instead of the pypi release, and that must build using it corresponding https://github.com/geopython/pywps/blob/main/environment.yml including dev/doc.

I can find the links this way:
```shell
docker run --rm birdhouse/finch:0.13.3-dev.3-with-chown-copy bash -lc 'python - <<"PY"
import glob, json, os
meta = "/opt/conda/envs/finch/conda-meta"
# Is sphinx installed?
sphinx = sorted(glob.glob(os.path.join(meta, "sphinx-*.json")))
print(f"sphinx_installed={bool(sphinx)}")
if sphinx:
    print("sphinx_pkg=", os.path.basename(sphinx[0]))
# Which installed packages declare sphinx as dependency?
parents = []
for fp in glob.glob(os.path.join(meta, "*.json")):
    with open(fp, "r", encoding="utf-8") as f:
        data = json.load(f)
    deps = data.get("depends", []) or []
    if any(d.split()[0] == "sphinx" for d in deps):
        parents.append((data.get("name"), data.get("version"), os.path.basename(fp)))
print("parents_of_sphinx:")
for p in sorted(parents):
    print(" -", p[0], p[1], p[2])
PY'
```
```shell
sphinx_installed=True
sphinx_pkg= sphinx-9.1.0-pyhd8ed1ab_0.json
parents_of_sphinx:
 - pywps 4.7.0 pywps-4.7.0-pyhd8ed1ab_0.json
 - sphinxcontrib-applehelp 2.0.0 sphinxcontrib-applehelp-2.0.0-pyhd8ed1ab_1.json
 - sphinxcontrib-devhelp 2.0.0 sphinxcontrib-devhelp-2.0.0-pyhd8ed1ab_1.json
 - sphinxcontrib-htmlhelp 2.1.0 sphinxcontrib-htmlhelp-2.1.0-pyhd8ed1ab_1.json
 - sphinxcontrib-qthelp 2.0.0 sphinxcontrib-qthelp-2.0.0-pyhd8ed1ab_1.json
 - sphinxcontrib-serializinghtml 2.0.0 sphinxcontrib-serializinghtml-2.0.0-pyhd8ed1ab_0.json
```

## Changes:

* Update Docker build to employ only core packages (removed dev/doc from ``environment.yml``).
* Optimized Dockerfile to avoid duplicating conda environment layer (and therefore duplicating size as well) when setting nonroot user.
* Fixed the `CHANGELOG.rst` rendering (not using double code quotes)

## Additional Information

@Zeitsperre @cehbrecht
I recommend removing the dev/doc references from pywps's `environment.yml` altogether to remove unnecessary packages at runtime for all users relying on conda-forge release.

In the meantime, a patch can be applied locally by doing:
```diff
dependencies:
  - python >=3.11,<3.15
  - pip >=26.1.0
- - pywps >=4.6
+ - pip:
+   - pywps >=4.6 
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,6 +138,11 @@ serialize = [
 ]
 
 [[tool.bumpversion.files]]
+filename = "Makefile"
+search = "APP_VERSION ?= \"{current_version}\""
+replace = "APP_VERSION ?= \"{new_version}\""
+
+[[tool.bumpversion.files]]
 filename = "src/finch/__version__.py"
 search = "__version__ = \"{current_version}\""
 replace = "__version__ = \"{new_version}\""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -272,7 +272,7 @@ override_SS05 = [
 addopts = [
   "--color=yes",
   "--strict-markers",
-  "--tb=native",
+  "--tb=native"
 ]
 python_files = ["test_*.py"]
 markers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -273,7 +273,6 @@ addopts = [
   "--color=yes",
   "--strict-markers",
   "--tb=native",
-  "-p no:flake8"  # avoid auto-detect, run flake8 separately to avoid incompatibility with pytest>=9 and pytest-flake8
 ]
 python_files = ["test_*.py"]
 markers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -273,7 +273,7 @@ addopts = [
   "--color=yes",
   "--strict-markers",
   "--tb=native",
-  "-p no:flake8"  # avoid auto-detect, run flake8 separately to avoid incompatibility with pytest>=9 and pytest-flake8
+  "-p no:flake8" # avoid auto-detect, run flake8 separately to avoid incompatibility with pytest>=9 and pytest-flake8
 ]
 python_files = ["test_*.py"]
 markers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,6 @@ dev = [
   "pylint >=3.3.3",
   "pytest >=8.0.0",
   "pytest-cov >=5.0.0",
-  "pytest-flake8",
   "ruff >=0.14.0",
   "vulture >=2.14",
   "watchdog >=4.0.0"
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -268,7 +272,8 @@ override_SS05 = [
 addopts = [
   "--color=yes",
   "--strict-markers",
-  "--tb=native"
+  "--tb=native",
+  "-p no:flake8" # avoid auto-detect, run flake8 separately to avoid incompatibility with pytest>=9 and pytest-flake8
```
Zeitsperre: Is this still required if `pytest-flake8` has been removed?

Zeitsperre (approved):

Zeitsperre:
Tests are failing due to server error at Ouranos. We're working on it.

pyproject.toml
fmigneault: I guess not needed. It can be removed. If some dev still has it locally, they would need to remove it manually as well.

fmigneault: removed in [`b6d64a4` (this PR)](https://github.com/bird-house/finch/pull/544/commits/b6d64a48e0d3f2d0979eb334dce796a3bea531ff)

PR #547, opened 2026-07-09
Allow min_members parameter use

tlogan2000:
## Overview

Changes:

Exposes a new input parameter `min_members` for ensemble processes, allowing users to specify the minimum number of ensemble members required for calculations. This allows users to ensure that results are only produced when a sufficient number of members are available, enhancing the reliability of the output.


Zeitsperre (approved):
LGTM

PR #549, opened 2026-08-04
Docker publishing

### LINKED ISSUE #550: Replace `lxml` in development deps or remove from conda env
This issue stems from a licensing issue.

`lxml` is available under GPLv2 and unfortunately, that doesn't work for us. While we don't include this in our Python pakcages, we do list this in the development (`dev`) install recipe.

For Python packaging, migrating to using `[dependency-groups]` (PEP 735) can bypass this problem for the `wheel` but our tests rely on this package. We currently install *all* dependencies in the Docker image (through the conda environment).

I can see a few ways forward:
- Create a stripped-down version of `environment.yml` that doesn't include development dependencies for the Docker image.
- Replace `lxml` with a similar XML parsing tool under a more permissive license (like https://github.com/mohammadraziei/pygixml).

Zeitsperre:
## Overview

Changes:

* Added a Docker build and Docker Hub publishing workflow.
* Removed the `addnab/docker-run-action` from `docker-testing.yml`; Now simply uses `bash` calls.
* Adjusted the Dockerfile metadata to use org.opencontainers metadata conventions and run commands for readability.
* Fixed the Zenodo DOI badge to use the latest URL.
* `bump-my-version` configuration is now more elaborate, changes to files on `main` will now bump version.
* Splits the `environment.yml` to separate development dependencies not needed in the docker image.
* Adds `deptry` to linting steps in order to audit all imported dependencies.

## Related Issue / Discussion

Integrates #551; closes #550

## Links to other issues or sources.

https://github.com/Ouranosinc/raven/pull/617
https://github.com/Ouranosinc/jupyterhub/pull/42

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,45 @@
+name: Docker Hub Publishing
+
+on:
+  push:
+    tags:
+      - "*"
+
+permissions:
+  contents: read
+
+env:
+  LATEST_TAG: birdhouse/finch:latest
+  RELEASE_TAG: birdhouse/finch:${{ github.ref_name }}
+
+jobs:
+  build:
+    name: Build and publish image on Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+        with:
+          disable-sudo: true
+          egress-policy: audit
+
+      - name: Checkout Repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+
+      - name: Build Docker image and publish to Docker Hub
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          push: true
+          tags: ${{ env.LATEST_TAG }},${{ env.RELEASE_TAG }}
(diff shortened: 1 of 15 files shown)


tlvu (approved):
LGTM, same comment as https://github.com/Ouranosinc/raven/pull/617#pullrequestreview-4876243730 ?

======
Same workflow as the https://github.com/Ouranosinc/jupyterhub/pull/42?

Meaning
* `bump-my-version {major | minor | patch | date | build}` or `bump-my-version bump --new-version {#.#.#-YYYYmmdd | #.#.#-YYYYmmdd.b}` support?
* if a tag is seen, the Docker image built on Github will also be pushed to DockerHub?

======

Zeitsperre:
> `bump-my-version {major | minor | patch | date | build}` or `bump-my-version bump --new-version {#.#.#-YYYYmmdd | #.#.#-YYYYmmdd.b}` support?

We don't use `date` here, so that's not supported. `--new-version` is supported but should almost never be required unless a maintainer has really botched something. Standard bump calls should be enough.

> if a tag is seen, the Docker image built on Github will also be pushed to DockerHub?

That's the plan, yeah.

pyproject.toml
```diff
@@ -137,19 +137,45 @@ serialize = [
...
+filename = "src/finch/__version__.py"
+parse = "__version__ = \"{current_version}\""
+replace = "{new_version}"
 
 [[tool.bumpversion.files]]
 filename = ".cruft.json"
-search = "\"version\": \"{current_version}\""
-replace = "\"version\": \"{new_version}\""
+parse = "\"version\": \"{current_version}\""
+replace = "{new_version}"
```
fmigneault: Curious about the case that was tripping up with `search`?
Usually, `parse` is used to define how to extract the parts (eg: `(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)`).

Zeitsperre: Odd, this shouldn't work according to the docs, but it seems to anyway? bug/feature? Will revert to using `search`.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Checkout Repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Gather Docker metadata
+        id: meta
+        uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6.2.0
+        with:
+          images: pavics/raven
```
fmigneault, line 41: @Zeitsperre 
Just saw this. Supposed to be `pavics/finch` or `birdhouse/finch`?

Zeitsperre: Yep, haha

PR #551, opened 2026-08-19
Drop lxml

### LINKED ISSUE #550: Replace `lxml` in development deps or remove from conda env

aulemahal:
## Overview

This PR fixes #550

Changes:

* Remove `lxml` as a dependency, replace with built-in  `xml.etree.ElementTree`. We were using two functions from `lxml`, without any of the added features.
* `xml.etree.ElementTree` is vulnerable to some well known attacks and `ruff` was warning about it. However, we only use it in the tests and with xml sources we generated ourselves. So I disabled the warnings.

## Additional Information
~However, `lxml` is dependency of `pywps`.  This PR makes it so WE don't _explicitly_ install it, but it is still present in the environment, the responsibility is shifted upstream.~

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -44,7 +44,6 @@ dependencies:
   - flake8-rst-docstrings >=0.4.0
   - flit >=3.11.0,<4.0
   - geojson
-  - lxml
   - nbsphinx >=0.9.8
   - nbval >=0.10.0
   - owslib
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,6 @@ dev = [
   "flake8-rst-docstrings >=0.4.0",
   "flit >=3.11.0,<4.0",
   "geojson",
-  "lxml",
   "matplotlib >=3.5.0",
   "nbconvert",
   "nbsphinx>=0.9.5",
@@ -323,7 +322,9 @@ ignore = [
   "COM", # commas
   "D205", # blank-line-after-summary
   "D400", # ends-in-period
-  "D401" # non-imperative-mood
+  "D401", # non-imperative-mood
+  "S314", # suspicious-xml-element-tree-usage
+  "S405" # suspicious-xml-etree-import
 ]
 preview = true
 select = [
--- a/tests/_common.py
+++ b/tests/_common.py
@@ -50,12 +50,12 @@ def get_output(doc):  # noqa: D103
     return output
 
 
-def get_metalinks(doc: xml.etree.ElementTree):
+def get_metalinks(doc: xml.etree.ElementTree.Element):
     """Return a dictionary of metaurls found in metalink XML, keyed by their file name.
 
     Parameters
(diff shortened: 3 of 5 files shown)


aulemahal:
Wait. @Zeitsperre , isn't lxml under BSD ?
https://github.com/lxml/lxml/blob/b946302cb9eb3e378150f69720ffdb40cc8c314f/LICENSE.txt

Zeitsperre:
Jesus. It is, but it looks like their docs aren't:
image

Zeitsperre:
@aulemahal This is still an improvement in all honesty. We rely far too much on some packages that provide one feature that can be rewritten using standard lib.

Zeitsperre (approved):

fmigneault:
I wasn't on the review, so could not flag sooner. Saw it through reference in https://github.com/bird-house/finch/pull/549

Why is it removed?
`lxml` is way more secure with default limits/restrictions and much more efficient at parsing than native `xml`. 

- https://lxml.de/FAQ.html#is-lxml-vulnerable-to-xml-bombs
- https://lxml.de/performance.html

The flagged `xml.etree.ElementTree` are precisely what `lxml` guards against.
The proper way is to use the following, not use the direct `ElementTree` reference.

```python
from lxml import etree
# Create a custom parser with security settings enforced
parser = etree.XMLParser(
    resolve_entities=False,  # Block local file inclusions (XXE)
    no_network=True,        # Block external network calls
    huge_tree=False         # Keep memory boundaries active
)

# Parse the untrusted data using your restricted parser
tree = etree.fromstring(untrusted_xml_data, parser=parser)
```

Zeitsperre:
> I wasn't on the review, so could not flag sooner. Saw it through reference in #549
> 
> Why is it removed? `lxml` is way more secure with default limits/restrictions and much more efficient at parsing than native `xml`.

I was under the impression from the FOSSA analysis that this library was under a different license.

@aulemahal Sorry for the blunder; I jumped the gun on this and will need to revert those changes.

aulemahal:
No issue. I would still point out that our use of `xml.etree.ElementTree` was safe as it was only ingesting xml from known and trusted sources. But indeed, if `lxml` is already there and usable, we should use it.

